PR #29, opened 2014-12-04
maintain a cache of opened files

chambart:
While testing some not so subtle program, I ended up receiving Sys_error("Cannot allocate memory").
The reason was that the maximal virtual memory limit was exceeded, because some files where read _a lot_. This patch fixed my problem.

Not that this is some kind of proof of concept and based on 1.2.0 (i'm using irmin 0.8.3 which enforce this version). If you consider adding this I can redo a clean patch.

### DIFF
--- a/lib/git_unix.ml
+++ b/lib/git_unix.ml
@@ -93,6 +93,59 @@ end
 
 module D = struct
 
+  module Read_file_cache : sig
+    val find : string -> Cstruct.t option
+    val add : string -> Cstruct.t -> unit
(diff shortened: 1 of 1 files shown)


samoht:
I've moved your cache directly in the FS module, and use it only when needed (ie. for immutable files): https://github.com/samoht/ocaml-git/commit/3c7a4644f1747c26aea4bd40901ba29e7164652e

chambart:
Thank you.

PR #49, opened 2015-01-10
Smart HTTP support

samoht:
It was more painful than expected, especially bounding `Lwt_io` channels and `Cohttp` bodies together. But it seems to work ...


samoht:
Note: currently support clone and fetch only, not push. Remaining patches coming tomorrow.

lib/unix/git_unix.ml
```diff
@@ -36,44 +36,150 @@ module M = struct
...
+        | `POST ->
+          let writer =
+            Cohttp_lwt_unix.Request.make_body_writer ~flush:true req oc
+          in
+          Lwt_io.make ~mode:Lwt_io.output ~close:(fun () -> Lwt_io.close oc)
+            (fun bytes off len ->
+               let chunk = Bytes.create len in
+               Lwt_bytes.blit_to_bytes bytes off chunk 0 len;
+               Cohttp_lwt_unix.Request.write_body writer chunk >>= fun () ->
+               return len)
```
avsm: Hm, I wonder if this would be easier with a Cohttp_lwt_cstruct backend instead?  @djs55 wanted this for the xapi socket backend too

samoht: not sure. the idea here is just to consider the body as a normal channel (and do the chunk encoding automatically). Actually the "hard" bit is `http_ic`.

avsm: ahh that's a nice API -- i'll think about if it would fit into either conduit or cohttp

samoht: it's also parsing/skipping the HTTP headers, so yea would fit into cohttp better. The only issue is that is seems not possible to start reading the response while the the post body is not completely sent. Not sure if it is an issue with cohttp or with the Git servers.

samoht: @avsm I think it could help if conduit `close` function didn't close both input and output channels.

samoht:
Need a new release of `cohttp` to support lwt 2.4.7 to fix the Travis scripts.

avsm:
will do shortly

PR #51, opened 2015-01-18
Override TMPDIR for Git_unix.write_file

vklquevs:
On a system where TMPDIR is on its own partition, `Lwt_unix.rename`
will fail with `Unix.EXDEV` (can't rename over different filesystems).

This puts the temp file in the same directory as the target file.
Looks like a side-effect of 94701e6

### DIFF
--- a/lib/unix/git_unix.ml
+++ b/lib/unix/git_unix.ml
@@ -258,10 +258,11 @@ module D = struct
     | len -> rwrite fd (Cstruct.to_bigarray b) 0 len
 
   let with_write_file file fn =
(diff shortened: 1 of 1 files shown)


samoht:
Thanks!

samoht:
After more though, I am not sure that's the right way to fix this. This will cause `.git/refs/` to have lots of short-lived temporary files which can be very confusing to the Git tools which tracks new branch creation/deletion. I'll make a new patch to expose `temp_dir` in the interface, although setting `Filename.set_temp_dir` should also be enough ...

PR #59, opened 2015-01-26
Reduce the default size of the LRU cache

samoht:


samoht:
This was causing the memory usage to go very high on some use-cases.

avsm:
can this be cut into a release?

samoht:
See https://github.com/ocaml/opam-repository/pull/3485

PR #66, opened 2015-02-08
Add User-Agent: ogit/$version header to smart HTTP

vklquevs:
I've been testing this on bitbucket.org, but it (understandably) doesn't speak git unless the user-agent contains at least "git/".


samoht:
Thanks!

PR #75, opened 2015-04-15
Modified to allow abbreviated SHA1 and to mimic git-ls-tree and git-cat-file

codinuum:
I have modified the original to achieve the following:
* to allow short SHA1s for specifying Git objects, and 
* to mimic git-ls-tree and git-cat-file.

The latter is implemented based on binary search for a SHA1 within an index file, as seen in Git.

The patches could be better by catching up with the evolution of the original; The actual modifications have been performed several months ago...

### DIFF
--- a/bin/ogit.ml
+++ b/bin/ogit.ml
@@ -290,13 +290,17 @@ let ls_tree = {
   man  = [];
   term =
   let recurse_flag = mk_flag ["r"] "Recurse into sub-trees." in
+  let show_tree_flag = 
+    mk_flag ["t"] "Show tree entries even when going to recurse them." 
+  in
+  let only_tree_flag = mk_flag ["d"] "Show only the named tree entry itself." in
   let oid =
     let doc = Arg.info [] ~docv:"SHA1"
         ~doc:"The SHA1 of the tree." 
     in
     Arg.(required & pos 0 (some string) None & doc ) 
   in
-  let ls (module S: Store.S) recurse_flag oid =
+  let ls (module S: Store.S) recurse_flag show_tree_flag only_tree_flag oid =
     run begin
       S.create () >>= fun t ->
 
@@ -306,11 +310,11 @@ let ls_tree = {
           | _       -> "blob",   false
         in
 
-        let rec walk recurse path sha1 =
+        let rec walk recurse show_tree only_tree path sha1 =
           S.read_exn t sha1 >>= fun v -> begin
             match v with
             | Value.Blob blob -> begin
-                printf "blob %s %s\n%!" (SHA.to_hex sha1) path;
+                printf "blob %s %s\n" (SHA.to_hex sha1) path;
                 return_unit
             end
             | Value.Tree tree -> begin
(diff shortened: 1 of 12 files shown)


bin/ogit.ml
```diff
@@ -177,6 +178,63 @@ let cat = {
       end in
     Term.(mk cat_file $ file)
 }
+*)
```
samoht: I personally find that command useful, so I'd like to keep if possible. So maybe simply rename it to `git cat`?

lib/FS.ml
```diff
@@ -215,6 +276,23 @@ module Make (IO: IO) = struct
           SHA.of_hex p
         ) packs in
       Lwt.return packs
+*)
```
samoht: it's better to read the diffs if you delete the previous code instead of commenting it out. So I'm not sure to understand what the `packs_opt` variable is for. What happens if someone runs `git gc` to create a new pack file behind your back? How do you update your cache of pack hashes?

lib/FS.ml
```diff
@@ -303,17 +401,59 @@ module Make (IO: IO) = struct
 
     let mem_in_pack t pack_sha1 sha1 =
       Log.debug "mem_in_pack %s:%s" (SHA.to_hex pack_sha1) (SHA.to_hex sha1);
+(*
       read_keys t pack_sha1 >>= fun keys ->
       Lwt.return (SHA.Set.mem sha1 keys)
+*)
```
samoht: same thing here, delete the old code instead of commenting it out.

lib/tree.ml
```diff
@@ -69,7 +69,7 @@ let string_of_perm = function
   | `Normal -> "100644"
   | `Exec   -> "100755"
   | `Link   -> "120000"
-  | `Dir    -> "40000"
+  | `Dir    -> "040000"
```
samoht: are you sure about this one? this means `perm_of_string (string_of_perm x)` becomes different of `x`.

samoht:
Thanks very much for your patches! I do appreciate that is not very easy to follow with the fast evolution of the project.

Few comments:

- the `cat-file` and `ls-tree` commands: ok, but I need the previous `cat-file` (which takes a filename instead of a hash) implementation to still be available. You can rename it to `cat` of `cat-raw` or something else.

- the objects to speed-up the pack file manipulation. I think that's a good idea, I'm fine to replace the previous code with that completely as it is a pain to keep track of the cached state anyway. But I'd prefer if the `class type` do not appear in the signature at all (ie. keep the type abstract and provide accessor functions)

- the short hashes function: I'd prefer to have a separate handling for this hashes (eg. can be either short or long) which a dedicated `read_file_short` function to be used by the `ogit` command-line tool. The rational is that you want fast reads when you are using `ocaml-git` as a library.

Thanks again!

samoht:
Thanks for the new patches, almost good to merge. Last point: can you run `ocp-indent` on your code to have the indentation consistent with the rest of the code?

codinuum:
I have run `ocp-indent` on modified files. I hope they are OK.
Note that the short hash function is not separated yet.

samoht:
Thanks! Do you plan to separate the short-hash functions?

samoht:
The patches look good to me, merging now. Thank you very much for your contribution!

codinuum:
I'm happy to hear that. But I'm sorry not to separate the short hash module from the patch.

The task was much more complicated than I expected...

lib/pack_index.ml
```diff
@@ -224,3 +224,312 @@ let add buf ?level:_ t =
...
+        (*Log.debug "prev_ok:%B next_ok:%B" prev_ok next_ok;*)
+        if prev_ok && next_ok then begin
+          let idx = idx_ofs + p in
+          (*Log.debug "c#scan_sha1s: idx -> %d (short)" idx;*)
+          raise (Idx_found idx)
+        end
+        else
+          raise SHA.Ambiguous
+      end
+      else if SHA.lt sha1 s then
```
samoht: @codinuum I'm reviewing that part of the code and I'm not sure to understand what you are doing here. The objects in the pack file are not ordered by SHA1 so I don't understand why this comparison would work.

samoht: Ha sorry the "spec" says that the keys are sorted. I've missed that...

PR #76, opened 2015-04-20
Constrain mirage-fs-unix to a version that has the connect fix or build ...

avsm:
...fails

### DIFF
--- a/opam
+++ b/opam
@@ -30,7 +30,7 @@ depends: [
   "lwt" {>= "2.4.7"}
   "hex"
   "alcotest" {test}
-  "mirage-fs-unix" {test}
+  "mirage-fs-unix" {test & >="1.1.4"}
   "mirage-types-lwt" {test}
   "cohttp" {test}
   "conduit" {test}


avsm:
This OPAM file generally seems wrong, since the tests are enabled by `"--%{mirage-fs-unix+alcotest:enable}%-tests"`, which might be activated without the `{test}` predicate (which the depenedencies use).

samoht:
Yea, we don't have a better way to enable/disable tests yet. This is tracked by https://github.com/ocaml/opam/issues/942

avsm:
but the issue is that you have marked test dependencies with `{test}`, but they are necessary with alcotest+mirage-fs-unix as well.

samoht:
@avsm I don't understand your last sentence.

Basically, the `test` dependencies are enable when you use `opam install <package> -t`. In that case, they are also depopts which means that we will compile the tests even if we don't have the `-t` but at least we will not run them. The situation is not ideal, but it works (I think). Do you have failing traces available?

avsm:
Yeah see mirage/irmin#195 for a failing travis test in imaplet-lwt.  In it, `git` is installed *after* `irmin` and so `irmin.unix` isnt available. That shouldn't happen with a normal depopt.

e.g. https://travis-ci.org/ocaml/opam-repository/jobs/59724112#L2032

```
  installed nocrypto.0.2.2
∗  installed x509.0.2.0
∗  installed irmin.0.9.1
Processing 53/59: [core_kernel: make] [tls: ocaml setup.ml]Thu Apr 23 14:25:11 UTC 2015 - building ...
∗  installed tls.0.2.0
∗  installed conduit.0.7.0
∗  installed git.1.4.3
∗  installed cohttp.0.17.0
∗  installed core_kernel.112.24.00
```

PR #80, opened 2015-06-05
Cohttp 0.18.0 compatibility

rgrinberg:
This is just quick fix in case you want to make a release that unbreaks
everyone.

@samoht I'm just curious, why does ocaml-git need such low level access to
cohttp internals anyway? I would like to understand the use case.


samoht:
The smart HTTP protocol is a bit tricky. Basically it mimics an RPC protocols, but all messages contains the whole history of the conversation (so the server doesn't have to have state).

PR #83, opened 2015-06-15
Fetch all remote refs on synchronize

AltGr:


samoht:
Why do you need this? Can it be chosen by passing an extra argument?

AltGr:
Didn't find it, but if there is one, that would be great!
At the moment it seems the refs are fetched by ogit, but not the commits they point to.

samoht:
ok, merging. The rational was to download only the commits you are interested in (ie. the one of the current branch) but it seems that the normal `git` command-line doesn't really like it. thanks for the patch!

AltGr:
Note that I am not sure it behaves correctly when there is no remote head yet

PR #87, opened 2015-06-18
Fix tree expansion to the filesystem.

samoht:


samoht:
This is related to https://github.com/mirage/irmin/issues/209

PR #93, opened 2015-07-02
Fix the fd leak and few other bugfixes and cleanups

samoht:
This should fix the virtual memory leak. @AltGr can you test it to see if that's better? Thanks!

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,4 +1,4 @@
 PKG cstruct dolog ocamlgraph re zip uri lwt mstruct cmdliner mirage-types
-PKG nocrypto hex cohttp.lwt mirage-flow tcpip mirage-http
+PKG nocrypto hex cohttp.lwt mirage-flow tcpip mirage-http alcotest
 B _build/**
 S lib/
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
-### 1.5.2
+### 1.5.2 (2015-07-04)
 
 * Fix handling of empty paths (#89)
+* Fix the serialization of dates in commit objects
+* Expose `Git.Packed_value.PIC.pretty`
+* Improve the efficiency of `Git_unix.FS.remove`
+* Support shallow packs (#81)
+* Fix an mmap leak introduced in `1.5.*` (#90)
+* Remove the dependency to OUnit for the tests
+* Improve the pretty printers and the output of `ogit`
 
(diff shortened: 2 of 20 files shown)


AltGr:
Quickly tested on samoht/master (21930ccb):
* `~unpack`: results fixed! It still takes up to 3GB, and fds jump to 200 during the last unpack stage (99%) (on raw plot)
* memory backend: fixed too! usage gets above 2GB (on opam-repository fetch) but it works.

The raw memory plot isn't very representative, I'll try some runs with ocp-memprof to get after-gc meterings.

Thanks :)

AltGr:
Forgot to mention the most important: no more virtual mem problem! Usage remains perfectly normal (~2x resident memory).

samoht:
The `~unpack` could probably be improved later. For the in-memory backend, it's hard to not need so much memory (as we need to store all the inflated objects), but maybe there is a way to keep less things ...

samoht:
Thanks for testing this, I'm merging and cutting a new release.

AltGr:
Arg! Sorry I didn't spot this before... there is now a _resident_ memory leak :/
Investigating...

AltGr:
weird: can't reproduce with ocp-memprof, where everything seems fine... but /proc/pid/status reports no freed memory. And the poor @camelus gets killed by OOM after handling a few PRs (or so it seems).

@samoht, are you using memory outside of the heap that ocp-memprof can't see ?

opam-ci-mem-new

cap3

http://memprof.typerex.org/user/home/lg/s/464527645bfe7266d83e8f3c716f1090/2015-07-04_06-17-39_69992320acf694c539a87c00fbe082aa_24386/index.html

@OCamlPro-BozmanCagdas @OCamlPro-Iguernlala @lefessan -- anything obvious I am missing ?

samoht:
arg

> @samoht, are you using memory outside of the heap that ocp-memprof can't see ?

Bigarray (via cstruct) maybe?

samoht:
Or maybe `io-page` allocation? Not sure how to track that ...

AltGr:
Note: it seems it was already there on 1.4 : same graph on my backport branch

opam-ci-mem

See the big difference on the x axis though!

AltGr:
One more note: even if using a new `GitStore.create` (without keeping any reference to the older one) for every treated PR, the issue persists.

PR #103, opened 2015-07-15
fix for 32-bit os

gregtatcam:


samoht:
That looks very good, and it is fixing expensive copies during sync which should speed-up `ogit clone` quite a lot.

gregtatcam:
Great!

PR #106, opened 2015-07-28
s/bits/bytes/ in docstring for read_exactly

yomimono:
Judging by git_mirage.ml and the channel API it calls, "bytes" is meant here rather than "bits".


samoht:
Hum seems that the last release of mirage-fs-unix breaks the unit tests. I'll merge your PR and I'll fix that, thanks!

PR #113, opened 2015-08-01
Better fetch

samoht:
Trying to simplify the Sync API. The current API is confusing as the difference from `fetch` and `pull` isn't clear.


samoht:
the REVDEPS error should be fixed by https://github.com/ocaml/opam-repository/pull/4579

PR #119, opened 2015-08-08
Compatibilty with cohttp 0.19.1

rgrinberg:
* I've simplified the definition of CLIENT. I'm not sure why it needs
Cohttp_lwt.Client at all if it's doing request/response processing by
itself :)

* write_footer has been added here because it shouldn't be part of
cohttp's api.

### DIFF
--- a/lib/http/git_http.ml
+++ b/lib/http/git_http.ml
@@ -19,16 +19,10 @@ let (>>=) = Lwt.bind
 module Log = Log.Make(struct let section = "http" end)
 
 module type CLIENT = sig
-  include Cohttp_lwt.Client
-    with type 'a IO.t = 'a Lwt.t               (* FIMXE in cohttp *)
-
-  module Request : Cohttp.S.Http_io with module IO = IO and type t = Cohttp.Request.t
-  module Response : Cohttp.S.Http_io with module IO = IO and type t = Cohttp.Response.t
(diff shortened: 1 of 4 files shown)


samoht:
Thanks! I'm not sure to understand why `write_header` is exposed in the API but not `write_footer` though :-)

rgrinberg:
Well `write_header` seems more sensible to me because it sends a request/response (sans the body) down the wire. While `write_footer` is basically just terminating a chunked body. I think that functionality just belongs in the body itself because it's specific to chunked encoding only.

dsheets:
If I recall correctly, there are such things as trailing headers ("trailers") in the chunked encoding. I believe a fully-featured `write_footer` would write those as well?

rgrinberg:
Yeah but I'd still argue this is a chunked encoding specific feature rather
than a component of a general http request/response.

samoht:
I agree that chunked encoding is maybe the only feature in HTTP which needs to add stuff in the HTTP header and have a footer in the body. Other features usually just add new HTTP headers. However a quick google search on HTTP footers gave me [that page](http://www.jmarshall.com/easy/http/) where it says:

> Footers are also rare, but might be appropriate for things like checksums or digital signatures.

So it might make sense to expose this in cohttp.

samoht:
(btw, the current footer I'm interested in is more the last blank line than the actuall HTTP footer)

Issue #125, opened 2015-08-29
ocaml-git is using too much memory


Chris00:
This correspond to the [following Irmin bug](https://github.com/mirage/irmin/issues/263).  Code (`c.ml`):

``` ocaml
open Lwt

module Store = Git_unix.FS
module G = Git_unix.Sync.Make(Store)

let main uri =
  Store.create ~root:"/tmp/c.dir" () >>= fun t ->
  G.fetch t uri ~update:true >>= fun res ->
  match Git_unix.Sync.Result.head_contents res with
   | Some h -> Store.write_head t h
   | None -> return_unit

let () =
  let uri = Git.Gri.of_string "git://github.com/xapi-project/xen-api.git" in
  Lwt_main.run(main uri)

;;
(* Local Variables: *)
(* compile-command: "ocamlfind ocamlopt -o c.native -package git.unix -linkpkg c.ml" *)
(* End: *)
```

Chris00:
ocaml-git 1.7.1

yallop:
I spent a while looking at this.  First, it's easily reproducible; `/usr/bin/time -v` reports

```
	Maximum resident set size (kbytes): 2401108
```

i.e. 2.3GB.

Digging about (using @kayceesrk's [allocation profiler](http://kcsrk.info/ocaml/profiling/2015/09/23/bytecode-allocation-profiler/)) reveals that most of the memory is allocated in the `Buffer` and `Bytes` modules.  Further investigation reveals that the [raw](https://github.com/mirage/ocaml-git/blob/e0ef97b8320cdaea54852c9533fe656a5b383a54/lib/packed_value.ml#L139) cache field of the `Packed_value.PIC.t` type is responsible for the majority of resident memory.  Removing the `raw` field altogether, and instead recomputing its value on each call to [unpack](https://github.com/mirage/ocaml-git/blob/e0ef97b8320cdaea54852c9533fe656a5b383a54/lib/packed_value.ml#L173-L179) brings the memory usage down significantly:

```
	Maximum resident set size (kbytes): 954948
```

i.e. 0.91GB.

Unfortunately, it also slows the program down significantly (at least on a machine with plenty of memory), from 5.42s user time to 9.06s.

[...]

PR #130, opened 2015-10-13
Improve logging performance: only build arguments when needed

yallop:
### Background 

Here's a typical piece of logging code with a simple performance problem:

```ocaml
Log.debug "received %s (%s)" (pretty k) (pretty v)
```

The problem is as follows: even if debug printing is turned off, the (possibly-expensive) function `pretty` is still called, although its output is discarded.

There are various approaches to avoiding the call to `pretty`, including

* syntax extensions (as in [Bolt](http://bolt.x9c.fr/)): we might introduce new syntax using Camlp4 that expands into code that is not executed when the logging level is too low:

  ```ocaml
    LOG "received %s (%s)" (pretty k) (pretty v) LEVEL DEBUG;
  ```

* a lazy wrapper around the logging call which is only forced when the logging level is sufficiently high

  ```ocaml
    Log.debug_force
     (lazy 
        (Log.debug "received %s (%s)" (pretty k) (pretty v)))
  ```

* a thunk around the logging call which is only called when the logging level is sufficiently high

  ```ocaml
    Log.debug_force
     (lazy 
        (Log.debug "received %s (%s)" (pretty k) (pretty v)))
  ```

* writing conditionals around each logging call

  ```ocaml
    if Log.debug_on () then
       (Log.debug "received %s (%s)" (pretty k) (pretty v))
  ```

None of these is entirely satisfactory.  Syntax extensions are rather heavyweight, and the other solutions involve an unfortunate repitition of the logging logic, since they all involve one call to decide whether to perform debug logging, and a second call to format and write the debug log entry.

An alternative solution, implemented in this pull request, is to change the logging interface to use CPS.  Rather than passing the arguments directly to `Log.debug`, we pass the context that expects the arguments to a thunk, and execute that thunk conditionally, as follows:

```ocaml
Log.debugk "received %s (%s)" @@ fun k ->
   k (pretty k) (pretty v)
```

This shares the advantage of the solutions above: the calls to `pretty` only occur if debug printing is enabled.  However, unlike those solutions there's no repetition of the logging level logic (and no need for a syntax extension).  The implementation of `debugk` is very simple; given a function `log` (e.g. [Dolog.log](https://github.com/UnixJunkie/dolog/blob/26e22e5b11f221623a72d9051a483937103213ba/lib/log.mli#L50)), we can write:

```ocaml
let debugk fmt k = 
  if Log.DEBUG >= get_log_level () then k (log level fmt)
```

### Performance improvements

This pull request adds CPS logging functions to the Dolog interface used in `ocaml-git` (3d6a0b8), switches the internal loggers over to the extended interface (37cccec) and rewrites debug logging calls that build arguments dynamically to use the new style (df8477b).  On the example #125 with the [optimised ocaml-hex](https://github.com/mirage/ocaml-hex/pull/13) implementation the new logging code results in 227MB less allocation overall.  With the unoptimised `ocaml-hex` the improvement is 320MB.

### DIFF
--- a/lib/FS.ml
+++ b/lib/FS.ml
@@ -22,7 +22,7 @@ let fail fmt = Printf.ksprintf failwith ("Git.FS." ^^ fmt)
 
 let err_not_found n k = fail "%s: %s not found" n k
 
-module LogMake = Log.Make
+module LogMake = Misc.Log_make
 
 module Log = LogMake(struct let section = "fs" end)
 
--- a/lib/FS.ml
+++ b/lib/FS.ml
@@ -212,7 +212,8 @@ module Make (IO: IO) (D: SHA.DIGEST) (I: Inflate.S) = struct
       | Some s -> some (Mstruct.to_string s)
 
     let read_aux name read_file t sha1 =
-      Log.debug "%s %s" name (SHA.to_hex sha1);
+      Log.debugk "%s %s" (fun log ->
+          log name (SHA.to_hex sha1));
       if SHA_IO.is_short sha1 then (
         Log.debug "read: short sha1";
         get_file t sha1 >>= function
@@ -301,7 +302,8 @@ module Make (IO: IO) (D: SHA.DIGEST) (I: Inflate.S) = struct
       LRU.clear keys_lru
 
     let read_pack_index t sha1 =
-      Log.debug "read_pack_index %s" (SHA.to_hex sha1);
+      Log.debugk "read_pack_index %s" (fun log ->
+          log (SHA.to_hex sha1));
       match LRU.find index_lru sha1 with
       | Some i -> Log.debug "read_pack_index cache hit!"; Lwt.return i
       | None ->
@@ -326,7 +328,8 @@ module Make (IO: IO) (D: SHA.DIGEST) (I: Inflate.S) = struct
         IO.write_file file ~temp_dir buf
 
     let read_keys t sha1 =
-      Log.debug "read_keys %s" (SHA.to_hex sha1);
+      Log.debugk "read_keys %s" (fun log ->
+          log (SHA.to_hex sha1));
       match LRU.find keys_lru sha1 with
       | Some ks -> Lwt.return ks
       | None    ->
@@ -347,13 +350,14 @@ module Make (IO: IO) (D: SHA.DIGEST) (I: Inflate.S) = struct
         IO.write_file file ~temp_dir pack
 
     let mem_in_pack t pack_sha1 sha1 =
-      Log.debug "mem_in_pack %s:%s" (SHA.to_hex pack_sha1) (SHA.to_hex sha1);
+      Log.debugk "mem_in_pack %s:%s" (fun log ->
+          log (SHA.to_hex pack_sha1) (SHA.to_hex sha1));
       read_pack_index t pack_sha1 >>= fun idx ->
       Lwt.return (Pack_index.mem idx sha1)
 
     let read_in_pack name pack_read ~read t pack_sha1 sha1 =
-      Log.debug "read_in_pack(%s) %s:%s" name
-        (SHA.to_hex pack_sha1) (SHA.to_hex sha1);
+      Log.debugk "read_in_pack(%s) %s:%s" (fun log ->
+          log name (SHA.to_hex pack_sha1) (SHA.to_hex sha1));
       read_pack_index t pack_sha1 >>= fun i ->
       let index = Pack_index.find_offset i in
       match index sha1 with
@@ -407,7 +411,8 @@ module Make (IO: IO) (D: SHA.DIGEST) (I: Inflate.S) = struct
     | Some v -> Value.Cache.add_inflated sha1 v; Some v
 
   let rec read t sha1 =
-    Log.debug "read %s" (SHA.to_hex sha1);
+    Log.debugk "read %s" (fun log ->
+        log (SHA.to_hex sha1));
     match Value.Cache.find sha1 with
     | Some v -> Lwt.return (Some v)
     | None   ->
@@ -421,7 +426,8 @@ module Make (IO: IO) (D: SHA.DIGEST) (I: Inflate.S) = struct
       cache_add sha1
 
   and read_inflated t sha1 =
-    Log.debug "read_inflated %s" (SHA.to_hex sha1);
+    Log.debugk "read_inflated %s" (fun log ->
+        log (SHA.to_hex sha1));
     match Value.Cache.find_inflated sha1 with
     | Some v -> Lwt.return (Some v)
     | None   ->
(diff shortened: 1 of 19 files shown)


dbuenzli:
Why not simply use `%a` formatting instructions and Format.ikfprintf ? Something like:

```ocaml

let Log.debug fmt = 
  let k _ = ...
  if Log.debug > then Format.ikfprintf k ppf fmt else Format.kprintf k ppf fmt

let pretty ppf v = ...
Log.debug "received %a (%a)" pretty k pretty v
```

Do you still get killed by the `fmt` parsing in `Format.ikfprintf` ?

yallop:
Thanks for the suggestion.  That'd work as well, and could perhaps be done with relatively few changes to the code by introducing some functions for turning functions that build strings into formatting functions:

```
let string_formatter f ppf x = output_string ppf (f x)
```

so that we could write

```ocaml
Log.debug "received %a (%a)" (string_formatter pretty) k (string_formatter pretty) v
```

I'll give it a try.

Drup:
I would argue that you should just have a pp function for every structure. `to_string` can then trivially be implemented in term of `pp`.

yallop:
I tried using `%a` instead of the CPS approach described above.  Here's the code:

https://github.com/mirage/ocaml-git/compare/master...yallop:logging-performance-format

Measuring shows that there's still some remaining allocation overhead, which adds up to about 50MB in the example in #125.  The overhead comes from traversing the structure representing the format string, which allocates a number of closures to receive the arguments, and a data structure to ensure that things happen in the right order.

Here are some more concrete measurements for do-nothing formatting with various numbers of arguments.

First, the CPS approach described in this pull request, i.e. code like this:

```ocaml
fprintfk stderr "foo %s %s"
        (fun log ->
           log
             (string_of_int x0)
             (string_of_int x1))
```

Here's the table of times for format strings with 0-7 `%s` arguments.  The running time is constant (and very low) and there's no allocation.

```
┌────────────────┬──────────┬─────────┐
│ Name           │ Time/Run │ mWd/Run │
├────────────────┼──────────┼─────────┤
│ k:0            │   4.49ns │         │
│ k:1            │   4.48ns │         │
│ k:2            │   4.66ns │         │
│ k:3            │   4.69ns │         │
│ k:4            │   4.69ns │         │
│ k:5            │   4.50ns │         │
│ k:6            │   4.48ns │         │
│ k:7            │   4.47ns │         │
└────────────────┴──────────┴─────────┘
```

Now here's the equivalent code written with `ifprintf` and `%a`, i.e. code of the form:


```ocaml
Printf.ifprintf stderr "foo %a %a"
        pp_int 0
        pp_int 1
```

Both the running time and the allocation increase roughly linearly with the number of arguments.  There's quite a bit of allocation:

```
┌────────────────┬──────────┬─────────┐
│ Name           │ Time/Run │ mWd/Run │
├────────────────┼──────────┼─────────┤
│ %a:0           │  10.31ns │   8.00w │
│ %a:1           │  21.71ns │  25.00w │
│ %a:2           │  53.09ns │  55.00w │
│ %a:3           │  83.35ns │  80.00w │
│ %a:4           │  90.58ns │ 105.00w │
│ %a:5           │ 109.22ns │ 130.00w │
│ %a:6           │ 133.28ns │ 155.00w │
│ %a:7           │ 156.19ns │ 180.00w │
└────────────────┴──────────┴─────────┘
```

yallop:
Here's an optimised version of the standard library `ifprintf`, which avoids allocating data structures and some other work:

https://github.com/ocaml/ocaml/compare/trunk...yallop:faster-ifprintf

It runs (very) approximately twice as fast as the unoptimised version and allocates (very) approximately half as much:

```
┌────────────────┬──────────┬─────────┐
│ Name           │ Time/Run │ mWd/Run │
├────────────────┼──────────┼─────────┤
│ %a optimised:0 │   9.78ns │   6.00w │
│ %a optimised:1 │  18.65ns │  16.00w │
│ %a optimised:2 │  31.17ns │  26.00w │
│ %a optimised:3 │  40.39ns │  36.00w │
│ %a optimised:4 │  55.42ns │  46.00w │
│ %a optimised:5 │  63.69ns │  56.00w │
│ %a optimised:6 │  81.22ns │  66.00w │
│ %a optimised:7 │  97.24ns │  76.00w │
└────────────────┴──────────┴─────────┘
```

Drup:
And with flambda ?

yallop:
Is there a version of flambda that works with core-bench?

Drup:
There are patches for core_bench here: https://github.com/OCamlPro/opam-flambda-repository-overlay

yallop:
Thanks.  I gave it a try, but the camlp4 package didn't build for me.  I'm not going to spend any more time on it, but it'd be interesting to hear whether flambda helps if anyone else manages to build it.

Drup:
After a bit of playing with flambda (and with @chambart to give me the right options)

1. In format, using `%a` is always (sometimes, much) faster than using `%d` (yes, it's counter intuitive).
2. When ignoring, format and printf are equivalent
3. When printing, format is twice slower than printf (indentation engine!)

 Now, as for what flambda brings: Let's consider the 3 competitors:

```ocaml
let log b fmt s k = if b then k (P.fprintf fmt s)
let with_printf_cont b fmt () =
  log b fmt "%d + %d = %d" @@ fun k -> k 1 2 3

let with_format_pp b fmt () =
  (if b then F.fprintf else F.ifprintf) fmt "%a + %a = %a" pp_int 1 pp_int 2 pp_int 3

let logf b fmt s k = if b then k (F.fprintf fmt s)
let with_format_pp_cont b fmt () =
  logf b fmt "%a + %a = %a" @@ fun k -> k pp_int 1 pp_int 2 pp_int 3
```

If `b` is *statically* known (like, there is a `let debug = false` somewhere), everyone is reduced to a noop. If `b` is behind a reference, in cycles/alloc: 

```
      printf_cont:    15.40    0
      format_pp:     241.09   82
      format_pp_cont: 14.16    0
```

My conclusion: Use the continuation based technique, with format, and use `%a` everywhere.

I didn't tried the special ifprintf by @yallop

dbuenzli:
@yallop & @Drup Thanks very much for the work, that's useful information.

Drup:
Oh, also, side conclusion : For the purpose of removing debug statement which presence is decided at compile time, syntax extension are not useful anymore: flambda will do it.

dbuenzli:
> syntax extension are not useful anymore: flambda will do it.

\o/

Drup:
#outofcontext :D

dbuenzli:
> #outofcontext :D

Ugh yes #truncatedquote.

samoht:
Can you rebase your patches to master? thanks!

yallop:
Rebased.

samoht:
@Magnuss there's still the same issue with jitsu and xenlight. Maybe a missing constraint somewhere?

MagnusS:
@samoht  `xenctrl` only installs `xenlight` if libxl 4.4 or newer is available. See also discussion here: https://github.com/mirage/jitsu/issues/29

The next version of Xenctrl (0.9.30) fails with an error when this happens, so it will be easier to debug. The PR is here: https://github.com/ocaml/opam-repository/pull/5015 (travis fails with precise, as expected :))

samoht:
Thanks!

UnixJunkie:
No volunteer to contribute such an improvement to the dolog library?
https://github.com/UnixJunkie/dolog

PR #133, opened 2016-01-20
Read files rather than mmapping them

talex5:
Otherwise, they don't get closed until GC time and we may fail with `ENFILE` or `EMFILE`.


samoht:
I'll fix the CI issue, fix a bug found by @justincormack and make a new release.

PR #136, opened 2016-02-26
git-unix: use larger buffer for large files

talex5:
Reading a 100 MB pack file went from 0.40s to 0.06s (there's still too much copying though - Python reads the same file in 0.05s).

### DIFF
--- a/lib/unix/git_unix.ml
+++ b/lib/unix/git_unix.ml
@@ -225,17 +225,16 @@ module IO_FS = struct
   let write_file file ?temp_dir b =
     with_write_file file ?temp_dir (fun fd -> write_cstruct fd b)
 
-  (* [read_into ~off buf ch] reads from [ch] into [buf] until
+  (* [read_into ~chunk_size ~off buf ch] reads from [ch] into [buf] until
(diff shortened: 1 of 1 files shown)


lib/unix/git_unix.ml
```diff
@@ -244,9 +243,14 @@ module IO_FS = struct
   let read_file file =
     Lwt_pool.use openfile_pool (fun () ->
         Log.info "Reading %s" file;
-        Lwt_io.(with_file ~mode:input) ~flags:[Unix.O_RDONLY] file (fun ch ->
+        Lwt_unix.stat file >>= fun stats ->
```
hannesm: maybe the LargeFile size? http://ocsigen.org/lwt/2.5.1/api/Lwt_unix.LargeFile otherwise 2GB limit..

talex5:
Since we return the file as a single Cstruct, we can't load files over 2GB on a 32-bit system anyway, so I don't think LargeFile is necessary (assuming regular stat can handle any size that fits in an OCaml int).

samoht:
Thanks!

PR #142, opened 2016-03-15
Use Logs and Fmt

samoht:


samoht:
This PR is not ready to merge as I have an issue with the reported (not investigated at all yet): the current one prints the name of the command-line tool, not the ones set-up in the various modules.

dbuenzli:
As written in the [documentation](http://erratique.ch/software/logs/doc/Logs_fmt.html#VALreporter) the default reporter that [you are using](https://github.com/mirage/ocaml-git/pull/142/files#diff-b03950bd348eefc89a0e8633441793e7R38) does not process or render information about message sources or tags. So you'll have to write your own reporter if you want to output information about sources.

Note that this interface will slightly [change](https://github.com/dbuenzli/logs/blob/3a77a94f641d14520527edac31da1381ee278acb/src/logs.mli#L314-L334) in the next version, see the future [release notes].(https://github.com/dbuenzli/logs/blob/3a77a94f641d14520527edac31da1381ee278acb/CHANGES.md).

samoht:
Argl, my `git push --force` deleted the comments :-/

lib_test/test_common.ml
```diff
@@ -17,37 +17,24 @@
...
-    begin match h with
-    | None -> ()
-    | Some h -> Fmt.pf ppf "[%a] " Fmt.(styled app_style string) h
-    end
-| Logs.Error -> pp ppf err_style (match h with None -> "ERROR" | Some h -> h)
-| Logs.Warning -> pp ppf warn_style (match h with None -> "WARN " | Some h -> h)
-| Logs.Info -> pp ppf info_style (match h with None -> "INFO " | Some h -> h)
-| Logs.Debug -> pp ppf debug_style (match h with None -> "DEBUG" | Some h -> h)
+let pad n x =
+  if String.length x > n then x else x ^ String.make (n - String.length x) ' '
```
dbuenzli: Doesn't something like: 

```ocaml
let pp_src ppf src = Format.pp_print_as ppf 10 (Logs.Src.name src) 
```
work ? (I don't know). Maybe we should have something for that in `Fmt`.

samoht: this doesn't seem to work. But maybe I'm doing something wrong... manual padding will go for now on.

PR #144, opened 2016-03-24
Add Store.size

talex5:
Before, the only way to get the size of a file was to decompress the whole thing.

The main win comes from getting the size of raw packed objects without decompressing. There are a couple of further improvements possible:

- Calculate the size of loose objects without loading them.

- For deltas in packs, only decompress the first few bytes (to get `result_size`).

This reduces the time to scan a large tree from 35s to 30s on my test repository. It's not a huge improvement, but probably the right thing to do. Still trying to work out what takes the rest of the time. For reference, Git manages to decompress and check out all the files into the working tree in 4.9s :-/


samoht:
Thanks!

PR #145, opened 2016-03-28
Add decompress in ocaml-git

dinosaure:
I don't know if you want a strong dependency with decompress or a weak dependency. I will do a release tomorow (just fix a bug with reverse stub in decompress, but this does not concern you :) ).


samoht:
Do you have some benchmarks? If perf hit is not too bad I'm happy to use it as the default and make camlzip an optional dependency.

samoht:
Can you also add the dependency to the `opam` file? Thanks!

dinosaure:
Hmmhmm, Decompress is only available on ocaml `>= 4.02.0` (`-safe-string`, `Bytes`, blabla). You want a version on `4.01.0` ?

samoht:
What's the status of that PR? I think I'd like to keep an optional dependency to Decompress for now on, as I'd like to be fully convinced on the robustness of Decompress first. Also, the performance hit is still noticeable, so it still makes sense to keep using zlib for a bit (I can easily be convinced to switch fully to Decompress at one point, though)

dinosaure:
Yes, I have lot of work for the moment (on MrMime) so you can close the PR and when I have the time to optimize Decompress, I come back with an optionnal dependency :+1: !

PR #146, opened 2016-04-04
Fix memory leak in IO.rec_files

samoht:
See https://github.com/ocsigen/lwt/issues/229

/cc @talex5

### DIFF
--- a/lib/unix/git_unix.ml
+++ b/lib/unix/git_unix.ml
@@ -186,17 +186,15 @@ module IO_FS = struct
     Lwt_pool.use mkdir_pool (fun () -> aux dirname)
 
   let list_files kind dir =
-    Lwt_pool.use openfile_pool (fun () ->
-        if Sys.file_exists dir then (
-          let s = Lwt_unix.files_of_directory dir in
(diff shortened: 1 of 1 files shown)


lib/unix/git_unix.ml
```diff
@@ -186,17 +186,15 @@ module IO_FS = struct
...
-      )
+    if Sys.file_exists dir && Sys.is_directory dir then
+      let d = Sys.readdir dir in
+      let d = Array.to_list d in
+      let d = List.map (Filename.concat dir) d in
+      let d = List.filter kind d in
+      let d = List.sort String.compare d in
+      Lwt.return d
+    else
+      Lwt.return_nil
```
rgrinberg: Should you run this `Lwt_preemptive.detach`?

samoht: I'm not very familiar with `detach`: does it deal with concurrent opened fd? Is it faster/slower/worth the complexity for this use-case?

rgrinberg: I see, I don't think it deals with concurrent fd's but it seems like `Sys.readdir` itself wouldn't be interrupted. But the only thing that you're trading off here is performance so I think what we have here is fine.

PR #147, opened 2016-04-06
Less unefficient Git_unix.read_file_with_read

chambart:
On some use case with a huge repository, this was seriously too slow. It was completely GC bound (doing lots of compactions).

This bring it back to reasonable performances.


samoht:
Thanks!

PR #160, opened 2016-09-29
Use safe-string

samoht:


lib/http/git_http.ml
```diff
@@ -61,8 +61,8 @@ module Flow(HTTP: CLIENT) (IC: CHAN) (OC: CHAN) = struct
           Lwt.return_unit)
       (fun buf off len ->
          let str = Bytes.create len in
-         Cstruct.blit_to_string buf off str 0 len;
-         HTTP.IO.write oc str >>= fun () ->
+         Cstruct.blit_to_bytes buf off str 0 len;
+         HTTP.IO.write oc (Bytes.to_string str) >>= fun () ->
```
yallop: The extra copy here is unfortunate.  Is it worth extending the Cohttp `IO` interface with bytes support?

samoht: @yallop but I am waiting forward to @seliopou's IO refactoring to avoid this.

PR #164, opened 2016-10-05
connect returns a value directly

yomimono:
Needed for mirage 3.


yomimono:
It's not clear to me why the 4.02 tests are failing.  I can build and run them locally, and it looks like Travis is installing a version of `git` from `opam-repository` instead of the package under test when it fails rather than the commit it's testing.

yomimono:
...it's because `PACKAGE="git.1.9.0"` is in the `global` section of `travis.yml`, I suspect.

PR #170, opened 2016-11-09
use a numerical version in META (_NUM strips leading v)

hannesm:


hannesm:
not urgent since currently you're using numeric tags (without leading "v")... but a good idea to have this uniform IMHO

PR #179, opened 2016-12-29
Use mirage-flow, mirage-channel and mirage-fs instead of mirage-types

samoht:


avsm:
merging this to get the tree compiling again -- can fix up revdeps in another PR

PR #186, opened 2017-02-14
Split the tests into `git-unix` and `git-mirage`

samoht:
The sync tests should be run on both backends before we can say the sync operations are "mature" enough (hint: this is not the case currently).


samoht:
The revdeps should be fixed by https://github.com/ocaml/opam-repository/pull/8477, so merging (there are other incoming PRs which will be testing the revdeps).

PR #210, opened 2017-05-11
Use Decompress

samoht:
Superseed #145 (and built on top of #209): the interesting commit is c544e3a

/cc @dinosaure

PR #211, opened 2017-05-12
Decompress (follow #210)

dinosaure:
Better implementation of `inflate`/`deflate` function. We allocate only one time the needed buffer (`input` and `output`) used to `inflate` __and__ `deflate`.  Because these functions is used in a _stop the world_ context, it's safe to use the same global buffer - otherwise, it's needed to have these buffer per serialization/deserialization contexts (and it's this case in `sirodepac`).

### DIFF
--- a/src/git/inflate.ml
+++ b/src/git/inflate.ml
@@ -34,59 +34,74 @@ module None = struct
 end
 
 module M = struct
-
   open Decompress
+
+  let size_buffer   = 0x8000
+  let input_buffer  = Bytes.create size_buffer
+  let output_buffer = Bytes.create size_buffer
+  let window = Window.create ~proof:B.proof_bytes
 
   exception Deflate_error of Deflate.error
 
-  let deflate ?(level = 4) data =
-    let input_buffer = Bytes.create 0xFFFF in
-    let output_buffer = Bytes.create 0xFFFF in
+  let deflate ?(level = 4) buff =
     let pos = ref 0 in
-    let res = Buffer.create (Cstruct.len data) in
-    Deflate.bytes input_buffer output_buffer (fun input_buffer -> function
-        | Some max ->
-          let n = min max (min 0xFFFF (Cstruct.len data - !pos)) in
-          Cstruct.blit_to_string data !pos input_buffer 0 n;
-          pos := !pos + n;
-          n
-        | None ->
-          let n = min 0xFFFF (Cstruct.len data - !pos) in
-          Cstruct.blit_to_string data !pos input_buffer 0 n;
-          pos := !pos + n;
-          n
-      ) (fun output_buffer len ->
-        Buffer.add_subbytes res output_buffer 0 len;
-        0xFFFF)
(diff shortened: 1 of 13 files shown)


samoht:
Force-pushed with a rebase on top of master

src/git/inflate.ml
```diff
@@ -33,6 +33,77 @@ module None = struct
...
+        let n = min size_buffer (Cstruct.len buff - !pos) in
+        Cstruct.blit_to_bytes buff !pos input_buffer 0 n;
+        pos := !pos + n;
+        n)
+      (fun output_buffer len ->
+      Buffer.add_subbytes res output_buffer 0 len;
+      size_buffer)
+      (Deflate.default ~proof:B.proof_bytes level)
+    |> function
+    | Ok _ -> Cstruct.of_string @@ Buffer.contents res
```
samoht: @dinosaure can we avoid a copy here?

src/git/inflate.ml
```diff
@@ -33,6 +33,77 @@ module None = struct
...
+        loop ~refill:(used_in t) (flush 0 size_buffer t)
+      | `Error _ -> None
+      | `End t ->
+        Mstruct.shift orig (used_in t - rest);
+
+        if used_out t <> 0
+        then begin
+          Buffer.add_subbytes res output_buffer 0 (used_out t);
+          Some (Mstruct.of_string (Buffer.contents res))
+        end else Some (Mstruct.of_string (Buffer.contents res))
```
samoht: Can we avoid a string->bigarray copy here? `Mstruct.of_cstruct (Cstruct.of_bigarry x)` is pretty cheap, would be better if we could use this.

src/git/inflate.ml
dinosaure: We can use a `Bigstring` instead a `Bytes` in `decompress` and use `Cstruct.of_bigarray` then in the best case. We can imagine, we allocate a `Cstruct.t` with `Cstruct.create (Cstruct.len buff)` as the result in the beginning (so, at most, the size of the input), but as I said we can do that in the _best case_ only. Sometimes, the result of `Deflate` can be upper than the input (when the input is too small and when we store the huffman tree and the header). Obviously, it's not a common case but I already seen that.

May be, a good way is to reimplement the `Buffer` module but with `Cstruct.t`. Another implementation is to start with a `Cstruct.t list`, reverse the list at the end and concat all `Cstruct.t`.

src/git/inflate.ml
dinosaure: We can optimize but the best best case, I think, is to have `Mstruct.pick_bigstring`.

dinosaure: Another point is to know exactly the size of the output (so `output_size <> None`).

src/git/inflate.ml
samoht: ok makes sense. Let's merge this one as is, we can optimise if needed later.

samoht:
All green, so merging!

PR #212, opened 2017-05-12
Mtime (last version 1.0.0 in OCaml 4.04.0)

dinosaure:


test/common/jbuild
```diff
@@ -3,4 +3,4 @@
 (library
   ((name      test)
    (wrapped   false)
-   (libraries (git mtime.os alcotest nocrypto lwt.unix logs.fmt))))
+   (libraries (git mtime.clock.os alcotest nocrypto lwt.unix logs.fmt))))
```
dbuenzli: You should also add `mtime` here.

samoht:
I've force-pushed to your branch with a rebase on top of master.

PR #214, opened 2017-05-12
Use digestif instead of nocrypto

samoht:
This removes the dependency towards libgmp and give some simple defaults to use
the library (e.g. Git.Mem is a full implementation of the Git protocol).


samoht:
Before merging this, need to check that this doesn't make the port to Xen and JavaScript more complicated. Normally it should be fine, but would be great to double-check.

samoht:
Decompress is now used

PR #220, opened 2017-06-19
Restore top level index

smeruelo:
I'm not sure if you have removed this index accidentally or on purpose, but now the link to the API from the README file is broken. Also, "up" navigation from git-unix/index.html and other pages is broken too.

So I've restored a part of the old index file (removed the part that links to the other removed indexes, like modules index, etc..). You may want to add it some style.


samoht:
The issue comes with the switch to `jbuilder` (the new build tool) and `odoc` (the new documentation tool). `odoc` likes to have a directory per library, but at the moment, it doesn't generate an top-level index: see https://github.com/janestreet/jbuilder/issues/63

So thanks for restoring the top-level index and hopefully this will be fixed before the next release so your change won't be overwritten but the release tool!

PR #225, opened 2017-07-25
Fix linking issues (which appears only 4.03 for some reasons)

samoht:


samoht:
Appveyor tests needs https://github.com/mirage/mirage-clock/pull/33 but seems otherwise fine

PR #227, opened 2017-07-25
New implementation of Git

dinosaure:
Hi all,

Tomorrow, I come back to Cambridge, then it's better to discuss about the next step of `ocaml-git` with a concrete code. So this PR will be the good place to talk about the performance, the API, the code and the kind of beer of the [Cambridge Blue](http://cambridge.pub/the-blue/).

To be clear, this PR need a strong acknowledge about what is Git. I don't explain what is a tree or a blob or the PACK file. I will do a conference [in India](https://confengine.com/functional-conf-2017/proposal/4052/git-under-the-hood-with-ocaml) for that.

## Wait, what is it ?

Some people from the mirage team know me, I started from February to improve `ocaml-git`, provide a new implementation of the PACK/IDX file, implement the `git gc` command, implement the `git push` and the HTTP protocol and some others stuff forgotten.

This work will (I hope) fix some bug about the memory consumption of `ocaml-git` too. The design of this library will be change internally for that. We will put some strong constraints about the memory consumption and change the scheme of the deserialization/serialization of a Git object.

For all of this, we need to rethink about the decoding of a Git object and, fortunately most of the Git objects are present on the PACK file. So, change the implementation of the decoding of the PACK file will change (I hope) this kind of bug.

However, when we write a new Git object, we write a loose Git object. So, it's the purpose of `git gc`. To avoid any fragmentation of the Git object inside your file-system, it's better to call cyclically a `compact` function. This function will PACK all of your Git objects inside one PACK file (and delete all others loose objects). So, we need to implement a smart _packer_. Before this implementation, `ocaml-git` just put the deflated Git object inside the PACK file and nothing else.

Now, this implementation come with a _delta-ification_ (I don't know the good word to explain this computation, but yes... deal with it) and compress the PACK file by the delta-ification. From some tests, we produce the same PACK file than `git` and sometimes, we produce a smaller PACK file than `git` - yes, we can be better than `git`!

So, about this compression and the possibility to store all objects in one PACK file, we can consider than we are close to implement the `git gc` command - now it's about some polishing computation like the memory needed to compact all (and avoid an explosion of your computer) and the deletion of _loose_ objects.

Finally, the PACK file is the key of the mystery of the chocolate. And the key of Git. Indeed, when you clone a Git repository, you receive a PACK file. When you push to a repository, you send a PACK file. So we need to take care about the implementation of the PACK file - I will explain precisely what is my implementation.

So for all of this, I refactored the `Sync` module and provide a better modular API to communicate (by the Smart protocol) with a remote Git repository. I re-implemented the clone command, I implemented a _negotiation_ engine (directly inspired by the `git` engine) to implement then the fetch command, and finally I implement the PUSH COMMAND!

To prove my work ALL COMMITS OF THIS PR WAS SEND BY MY IMPLEMENTATION!

Finally, we can restart the implementation of the HTTP protocol in the good way. So if you read all, you can say: oh wait, but you missed some goals of what we expected! Like, yes I did not yet implemented the HTTP protocol and yes, the `git gc` is not yet available. But I will start a detailed explanation of what happen for this PR, my choices, and why I took a long time to do all.

## [`sirodepac`](https://github.com/dinosaure/sirodepac)

Firstly, I started to implement only the PACK file in another project. The name is `sirodepac`, a famous lemon syrup in south of France. This project was focused only on the PACK file and try to decode a PACK file and encode a PACK file. But, because @avsm commits faster than its shadow, if you want to decode naively a big PACK file from any repositories of @avsm, your computer will die.

So, it's time to explain the strong constraints about my implementation:
* First, my decoder (and encoder) need to be a _non-blocking_ decoder. That means, we can stop at any step of the decoding, do what we want, and restart the process only with a `state` which we had kept.
* Secondly, the buffer needed to store the input must be of fixed size.
* Finally, any internals buffers has the same constraint

The last two point is very important because, in Git, we can handle some huge over big files (_blob_). And it's not realistic to store entirely the PACK file in the memory, compute some Git objects in the memory and have some other allocated data.

I wrote [a little article](https://dinosaure.github.io/sirodepac.html) (in French, sorry) about that and say something like:
* The state must be pure to be located only on the minor heap
* The fixed size buffer will be in the major heap but this object will live as long as we want to handle any Git objects.

### Decoder

Because the state will be in the minor heap, it will be deleted quickly. So the final point is, the decoding of the PACK file **never** allocates any buffer in the major heap. Only the client is able to allocate what is needed to decode a PACK file. In the truth, we allocate a little buffer of `0x7F` bytes, but only one, it's ok...

However, we have the _delta-ification_. The _delta-ification_ is:
* We have a _base_, it's a Git object
* We have a list of `hunk`
* And from the base and the `hunks`, we can construct a new Git object which is a _delta-ified_ object.

So, we need a second pass to construct all _delta-ified_ objects or we need the totality of the PACK file to resolve all _delta-ified_ objects. I provided these two kind of way to reconstruct any Git objects from a PACK file.

But, again, it's about the memory consumption. If we want to construct the new Git object, we need to keep the list of `hunk` which contains some little buffers (of, in maximum, `0x7F` bytes), we need to store the base Git object and a result buffer to store the result of the resolving. And, another point is a base Git object can be _delta-ified_ too (we can have a chain of 50 _delta-ification_ in the current implementation of Git).

You can understand that, if you implement this naively, your computer will burn.

So, my implementation is like a toolbox to deserialize any Git object and try to avoid any allocation of the memory. If you know the size of the biggest Git object of your PACK file, and if you know the deeper chain of the _delta-ification_ of your PACK file, you can allocate strictly what is needed to decode any object of this PACK file (_delta-ification_, or not). It's an unpleasant game between all of your buffers to store any information (_meta-data_, list of _hunks_, Git base object) strictly needed to construct the Git object requested. In other way, if you don't know one of this information requested, my decoder will allocate what is needed.

So, for all of that, I think, my implementation will fix the bug about the memory consumption because we can know strictly what is needed to construct any Git objects. However, to obtain the requested information, you need to pass one time on the PACK file and do some computation and this is cost.

If you want a concrete example of all of this, you can look the `clone` function which calculates this information, allocate what is needed, checks all Git object, and if it's a _thin-pack_, repack all of these Git objects to a new PACK file.

### Encoder

I think it's the more interesting part of my job, try to encode any Git object to a new PACK file and calculate a good application of the _delta-ification_. But the specification about this hard compute is only described in an IRC discussion between one guy and Linus Torvalds... I known better than that about the specification.

But, with @samoht, we try lot of time to have a good result and try to be close to the implementation of `git` - to produce the same small PACK file.

The key about the implementation of the _delta-ification_ is about the heuristic of which object is better as base to apply a _delta-ification_ to the current object. I don't want to explain the detail about this heuristic (we can read the IRC discussion!) but when we find a good base object, we will compute the delta-ification.

Before, I believe `git` do a patience diff between the source and the target to produce a good list of `hunks`. I re-implemented (you can look in the history of [the `sirodepac` repository](https://github.com/dinosaure/sirodepac/commit/fb9bf0479cee356d9b3aadea5535d6b867254851)) the [patience diff without Core](https://github.com/janestreet/patdiff). But the result was bad.

In truth, `git` uses a tuned implementation of the Rabin's fingerprint from [`libXdiff`](https://github.com/git/git/blob/master/xdiff/xdiffi.c) to generate a good list of `hunk`. And because the only reference implementation of this algorithm is in C, I put a C stub in this library (`rabin_stubs.c`) to calculate the Rabin's fingerprint. Then, I translated the tuned implementation from `git` to an OCaml code. I think, we can move this code to an OCaml code but I did not have the time to do this - and I think, it's not easy to do the translation because is very close to the representation of the integer in C.

But, finally, we produced a good PACK file and, as I said, sometimes we produce a better PACK file than `git` but for some others reasons.

And, from this _delta-ification_, you need to re-order the list of your Git objects (a topological sort) because Git has some assertion about the format of the PACK file (only understandably on the only exhaustive specification of Git, the C code).

Finally, we have an encoder (in the same way as the decoder) to encode this list to a PACK file. To be clear, at this time we use a lightweight representation of the Git object. You can understand than it's not possible to manipulate like 4 000 000 commits (and tree, and blob, and tag, ...) in your memory.

### Conclusion

We have a good implementation of the PACK file. It's not perfect, for example, we can optimize again the serialization (in the same way as `git`) and we can, again, restrict the memory consumption. But, I think, it's a good bedrock for the next things.

## `ocaml-git`

Then, the job is to integrate this big part in `ocaml-git`. And you can understand, it's not really easy.

### A bottom-to-top development

When you know the code-base of `ocaml-git` you can understand than why we have some bug about the memory consumption. Any decoder was not think in the _non-blocking_ way, they a use the totality (stored in the memory) of the input (and they need to have the totality of the input, because, exactly, the decoding is not _non-blocking_), and it's same about the serialization.

One other problem is to use an interface close to `camlzip` which one can not permit to be used to a _non-blocking_ decoder/encoder.

For all of these, it's needed to start a big replacement of some specific layer of `ocaml-git`. But, it's not like, this project is bad and we restart a new one. This project is good and the API is good and it's why it's a _bottom-to_top_ development, the goal is to provide the same previous API but with a big internal change.

So, I don't restart the project, I took it and try to improve some deeps layers - and, of course, this surgical strike changed a little bit the API.

### [Angstrom](https://github.com/inhabitedtype/angstrom)

I think, all people want to read this PR known `angstrom` from @seliopou. This is a good library and, I think, it's the best choice I did to handle the deserialization. In fact, it's easy to describe the Git format firstly and when you understand the problem of the alteration (and the problem to keep an internal buffer to not remove what was consumed by the alteration), you can provide a useful API to decode _everything_.

But, to be clear, and I explained this in the code,  all decoder/encoder stuff in this project is close to the Git format and because I know when we can have an alteration and how many bytes I need to keep while the alteration, I produced a _user-friendly_ interface. But, don't copy/past...

So the big deal is happen in the `Helper` module which one makes a new _non-blocking_ memory limited decoder from an Angstrom decoder.

### [Faraday](https://github.com/inhabitedtype/faraday) / [Farfadet](https://github.com/oklm-wsh/Farfadet)

However, I will not say the same for `Faraday` (sorry @seliopou). `Faraday` was think to be run in two cooperative processes, one to feed the encoder, the second to flush. But Faraday don't care about the memory consumption of the encoder because, in this pattern, we can say easily than the encoder will not growth.

So, I provided a Faraday encoder (make by the Farfadet library) but I don't use it. I still open about this question.

### Minienc

And yes, if I don't use Faraday, what I use to encode a Git object? My little pony^Wminienc. This module provide a memory limited encoder in the same way than Faraday (with the same interface). The big change is:

* If we are close to fulfill the internal buffer, we ask to the client to flush.
* The state is pure
* The internal Queue is pure (and is come from containers from @companion-cube and Chris Okasaki)

From all of this, I provided exactly the same interface. I did not have the time to play with GADT and provide a same library like `Farfadet`, but it still in my TODO.

Finally, from this `Minienc`, I can digest a fixed-size buffer to obtain the hash of a Git object.

### `camlzip` and `decompress`

I keep this compatibility from the previous version to use `camlzip` instead `decompress`. You need to know that is not easy to find a common interface between these two projects and I'm not sure if all work (but I prefer to believe than all is ok).

So, we can continue to improve the compression or the decompression of the Git object independently than `decompress`. And yes, the interface required is close to the interface of `decompress` but if a can provide the same interface with `camlzip` with the same internal constraint (the _non-blocking_ state), I think it's a good interface. Then, this interface is an interface "à la @dbuenzli" and, to use this kind of intf, I think it's a good choice to constraint the _inflater_ and the _deflater_ to this intf.

However, about the memory consumption, we can have a problem when we want to deflate. Indeed, for both,  the implementation need to keep all of the input to produce then an Huffman tree. For a large file, `git` calls `zlib` with a specific `FLUSH` method (like `SYNC_FLUSH`) to make a new deflate block and free new memory.

In the state, I don't do this and it can be a problem for a large file.

### CRC-32

I provide a C implementation of the CRC-32 checksum. For sure, if you want to have a pure OCaml implementation, I can but, I let this because it's not the biggest goal for the next step in my TODO list.

### FileSystem and Path

I decided to split the representation of the Path from the FileSystem because I consider than the Path can be useful to represent a reference independently than a file-system back-end (like for `Git.Mem`). Then, the `Store` module needs an implementation of the file-system which one provides a partial `read` and a partial `write` (to connect with a _non-blocking_ encoder/decoder) and the `mmap` function for the PACK file decoder (and resolve a delta-ified object in one pass).

However, this API can not be changed for sure and it's one other reason of why I decided to do a development _bottom-to-top_. Before this version, `ocaml-git` loaded all of the file in one time in the memory and, obviously, it's another point to explain the memory consumption of `ocaml-git`.

### [`Digestif`](https://github.com/mirage/digestif) and `Bytes.t`

I don't know what people think about that but for me, an end-user will use a lot of hash values. And for this reason, I consider than it's better to localized the value of hash in the minor heap as a `Bytes.t` instead a `bigstring`.

So I constrained in the PACK decoder/encoder than the result of the `digest` function will be a `Bytes.t`. I'm not sure about this choice and **we can change** (not easily on the other hand). But I think, we need a proof than it's better to use `bigstring` instead `Bytes.t`.

### Smart and Sync

And for all of this, I re-implemented, in the same way than `ocaml-imap` the Smart protocol.

I think, it's better to implement the decoder and the encoder in this way instead to use Angstrom but I explained why in the file.

Obviously, the API is more modular than the previous implementation of `clone`, `push` and `fetch`. In this API, we can change the negotiator engine, the packer engine and let the user to specify what he wants. The `main.ml` is a good example to understand how to use the `Sync` module. The implementation to connect, send and receive the date from the network is functorized - and in the example, you can look 2 implementation of the `Net` interface (one for `ssh`, the other to communicate with a `git` server).

I will push a documentation about `Sync` soon.

## What is the next?

As I said, I tried to do a _bottom-to-top_ development and if you know the API of the `Store` module, you can say than this work is not done yet (but close).

So the next, is to continue the development to provide the same previous interface. But I think, it's a good moment to have a critic about the API and make a better library to interact with a Git repository. From this project, we will obviously break the compatibility so, for me, it's an opportunity to make the best choice the next thing.

Voili, Voilou!


samoht:
Thanks! I will try to review all the patches shortly, but first a few question/remarks:

- about digestif/camlzip: I don't have a strong opinion about keeping the choice to have both (it was useful when digestif was not yet stable -- which is not the case anymore). If you thing you can do simpler and better things by breaking the compat with camlzip, please do: having a portable non-blocking codec for pack files is a good tradeoff vs. pure efficiency.

- about hashes in the major/minor heap: I agree that we need some perf numbers. Nocrypto uses cstruct buffers everywhere, so I guess @pqwy and @hannesm have already considered the question.

- about crc32: have you seen https://github.com/xapi-project/ocaml-crc? ocaml-git were using that before switching to a pure version of it, to ease portability.

- and more generally: it would be great to minimize the amount of C code that we add to this library, as it will cause headache when 1/ porting to Xen and 2/ porting to JavaScript. I could see that the parts where this is critical for speed could have two backends, but in general I would prefer the simplicity of having only one code portable code path written in OCaml. But some numbers would probably be needed there too.

hannesm:
crc32: please keep in mind that the xapi-project/crc is C code, and requires effort to get it running on all platforms

performance: I'd be happy to see some numbers (I don't have any) -- esp. whether a bytes/string backed `cstruct` is faster/slower than the current one (which uses bigarray).

samoht:
To ease the review, could you replace the existing code with the new one instead of adding a new subdirectory? Would be easier to catch API changes and what is currently missing.

dinosaure:
About CRC-32, I took the previous implementation available in `ocaml-git` 👍 .

About `decompress`/`camlzip`, I will keep the current interface. Indeed, we don't have any overhead with this interface and `decompress` (because, as I said, this interface is close to the interface of `decompress`). The available implementation of the `inflate`/`deflate` modules with `camlzip` was used to compare the ratio between `decompress` and `camlzip` and it seems than `decompress` as a good ratio.

Note: A good point to debug, `decompress` put in any way an empty flush at the end of the compression block, so when you want to extract a git object from a PACK file: the block starts with `78 9c` (the header) and finishes with `00 00 ff ff` - this trick saved my brain.

About `Digestif` and the constraint with the `Bytes.t`, it's because I need a concrete type to define a `get` function needed by the `Radix` functor. The `get : t -> int -> char` function is not available in the `IDIGEST` interface because it's not the purpose of an abstract hash representation (you should never use `get` on your hash). So, when I constraint the type of `Bytes.t`, then I can say in the lower layer than to get a character, I will use `Bytes.get`.

We have different way to fix this:
* We can consider than it's worth to put in the `IDIGEST` interface the `get` function
* We can put a proof (GADT "à la decompress") to know the concrete type of the hash and, by this proof, provide the function for the Radix functor
* We can keep this current structure

For the first and the second way, it will be easy to switch between a `bigstring` or a `Bytes.t` representation. For the current structure, we have lot of noise to change the concrete type (like change all constraint) - and it's why it's not easy to switch.

I will take my flight :) !

src/git/main.ml
```diff
@@ -0,0 +1,188 @@
+let () = Printexc.record_backtrace true
```
samoht: this should be merged with `src/git-unix/ogit/main.ml`

src/git/mem.ml
```diff
@@ -1,228 +0,0 @@
-(*
```
samoht: where did this go?

src/git/negociator.ml
```diff
@@ -0,0 +1,413 @@
...
+
+    go rev
+
+  type state =
+    { ready    : bool (* got ready detail *)
+    ; continue : bool (* got continue detail *)
+    ; finish   : bool (* want to finish *)
+    ; count    : int  (* how many hashes we get *)
+    ; flush    : int  (* how many hashes before flush *)
+    ; vain     : int  (* how many vainly hashes we have *)
```
samoht: what is a vain hash?

src/git/object_graph.ml
```diff
@@ -1,161 +0,0 @@
-(*
```
samoht: where did this go?

src/git/pack.ml
```diff
@@ -1,447 +1,1583 @@
...
-  let compare = compare
-  let equal = (=)
-
-  let pp ppf t =
-    Format.fprintf ppf "@[";
-    List.iter (fun p ->
-        Format.fprintf ppf "@[%a@;@]" Packed_value.PIC.pp p
-      ) t;
-    Format.fprintf ppf "@]"
+let pp_list ?(sep = (fun fmt () -> ()) )pp_data fmt lst =
```
samoht: Use `Fmt.list`

src/git/index.ml
```diff
@@ -1,345 +0,0 @@
-(*
```
samoht: This was implementing the handling of `git checkout`. Have you re-implemented that as well?

src/git/packedRefs.ml
```diff
@@ -0,0 +1,226 @@
+module type S =
```
samoht: please name this `packed_refs.ml`

src/git/common.ml
```diff
@@ -0,0 +1,119 @@
+module type BASE =
```
samoht: please rename this to `s.ml`

src/git/search.ml
```diff
@@ -1,93 +0,0 @@
-(*
```
samoht: would be great to re-add this.

src/git/smart.mli
```diff
@@ -0,0 +1,616 @@
...
+    constraint the {!IDIGEST} module to generate a {!Bytes.t}. *)
+
+module type CLIENT =
+sig
+  module Digest
+    : Ihash.IDIGEST with type t = Bytes.t
+  (** The [Digest] module used to make the module. *)
+
+  module Decoder
+    : DECODER with type Hash.t = Digest.t and module Digest = Digest
```
samoht: why does the DECODER needs to care about the Digest implementation?

src/git/store.mli
```diff
@@ -0,0 +1,737 @@
...
+
+      As {!read_p}, {!size_p} can return an {!error}.
+  *)
+
+  val size_s : state -> Hash.t -> (int64, error) result Lwt.t
+  (** [size_s state hash] is the same process than {!size_p} but we use the
+      state-defined buffer. That means the client can not use this function in a
+      concurrency context with the same [state]. *)
+
+  val size : state -> Hash.t -> (int64, error) result Lwt.t
```
samoht: I think I would just expose the top-level function (with the `_s` suffix) and move the non-blocking one (with the `_p` suffix into a sub-module).

src/git/store.ml
```diff
@@ -1,49 +1,1155 @@
...
+  val raw_s : t -> Hash.t -> ([ `Commit | `Tree | `Tag | `Blob ] * Cstruct.t) option Lwt.t
+  val raw : t -> Hash.t -> ([ `Commit | `Tree | `Tag | `Blob ] * Cstruct.t) option Lwt.t
+  val contents : t -> ((Hash.t * Value.t) list, error) result Lwt.t
+  val buffer_window : t -> Inflate.window
+  val buffer_zl : t -> Cstruct.t
+  val buffer_de : t -> Cstruct.t
+  val buffer_io : t -> Cstruct.t
+  val fold : t -> ('a -> ?name:Path.t -> length:int64 -> Hash.t -> Value.t -> 'a Lwt.t) -> path:Path.t -> 'a -> Hash.t -> 'a Lwt.t
+
+  module Ref :
```
samoht: how do you read/write references now? what about the test_and_set function?

src/git/tree.mli
```diff
@@ -1,38 +1,78 @@
...
+    ; node : Hash.t }
+  and perm =
+    [ `Normal (** A {!Blob.t}. *)
+    | `Everybody
+    | `Exec   (** An executable. *)
+    | `Link   (** A {!Blob.t} that specifies the path of a {i symlink}. *)
+    | `Dir    (** A sub-{!Tree.t}. *)
+    | `Commit (** A sub-module ({!Commit.t}). *)
+    ]
+  and t = entry list
```
samoht: The entry list is supposed to stay sorted (with a weird lexicographic order distinguishing tree and objects). Do you manage this? If yes, I guess the type `t` would be better be `and t = private entry list` to avoid random orders.

src/git/_tags
```diff
@@ -0,0 +1,4 @@
+true: safe_string, bin_annot, debug, keep_locs, use_menhir
+<*/**>: warn(@20@25@26@33@39)
+
+<*/**.{ml,mli,byte,native}>: package(fpath cstruct lwt lwt.unix angstrom faraday farfadet digestif.c bos decompress lru camlzip uri astring), linkdep(rabin_stubs.o)
```
dinosaure: Yes, it's to compile `main.ml`, I will delete this - for sure, the library don't have these dependence (`bos` was no longer necessary, `lwt.unix` is for `lwt_fs_unix.ml` and `digestif.c` is for `sha1.ml`.

src/git/minienc.ml
```diff
@@ -114,24 +114,17 @@ struct
...
-    let rec aux = function
-      | [] -> ()
-      | [ x ] -> pp_data fmt x
-      | x :: r -> Format.fprintf fmt "%a;@ " pp_data x; aux r
-    in
-
-    aux lst
+  let pp ppv ppf q =
+    Fmt.pf ppf "[ %a ]"
+    (Fmt.hvbox (Fmt.list ~sep:(fun ppf () -> Fmt.pf ppf ";@ ") ppv)) (to_list q)
```
samoht: `Fmt.l(ist ~sep:(unit "; @") ppv)`

src/git/negociator.ml
```diff
@@ -50,7 +41,7 @@ struct
         then [ "ALTERNATE" ] else [] ]
       |> List.concat
     in
-    pp_list ~sep:(fun fmt () -> Format.fprintf fmt " |@ ") Format.pp_print_string fmt flags
+    Fmt.list ~sep:(fun ppf () -> Fmt.pf ppf " |@ ") Fmt.string ppf flags
```
samoht: Same here: `Fmt.(list ~sep:(unit " |@ ") string)`

src/git/pack.ml
```diff
@@ -295,41 +274,47 @@ struct
...
+    | Insert k ->
+      Fmt.pf ppf "(Insert #k)"
+    | Copy k ->
+      Fmt.pf ppf "(Copy #k)"
+    | End ->
+      Fmt.pf ppf "End"
+    | Consume ->
+      Fmt.pf ppf "Consume"
+    | Exception err ->
+      Fmt.pf ppf "(Error %a)" (Fmt.hvbox pp_error) err
```
samoht: too many endlines, please keep the code compact :p

src/git/radix.ml
```diff
@@ -210,7 +200,8 @@ struct
 
   let to_list t = fold (fun (k, v) acc -> (k, v) :: acc) [] t
 
-  let pp pp_data fmt radix =
-    pp fmt "[ @[<hov>%a@] ]"
-      (pp_list ~sep:(fun fmt () -> pp fmt ";@ ") (fun fmt (k, v) -> pp_data fmt k v)) (to_list radix)
+  let pp ppv ppf radix =
+    Fmt.pf ppf "[ %a ]"
+      (Fmt.hvbox (Fmt.list ~sep:(fun ppf () -> Fmt.pf ppf ";@ ") (fun ppf x -> ppv ppf x)))
```
samoht: `Fmt.(list ~sep:(unit ";@ ") ppv)`

src/git/smart.ml
```diff
@@ -707,20 +696,20 @@ struct
     ; refs         : (Hash.t * string * bool) list
     ; capabilities : capability list }
 
-  let pp_advertised_refs fmt { shallow; refs; capabilities; } =
-    let sep fmt () = Format.fprintf fmt ";@ " in
-    let pp_ref fmt (hash, refname, peeled) =
+  let pp_advertised_refs ppf { shallow; refs; capabilities; } =
+    let sep ppf () = Fmt.pf ppf ";@ " in
```
samoht: `let sep = Fmt.unit ";@ "`

src/git/smart.ml
```diff
@@ -972,33 +961,15 @@ struct
...
-    let pp_unit fmt ()   = Format.fprintf fmt "()" in
-    let pp_string fmt s  = Format.fprintf fmt "%S" s in
-    let sep fmt ()       = Format.fprintf fmt ";@ " in
-    let pp_refname fmt r = Format.fprintf fmt "%s" r in
-    let pp_pair pp_a pp_b fmt (a, b) =
-      Format.fprintf fmt "(@[<hov>[@<hov>%a], @[<hov>%a@]@])"
-        pp_a a pp_b b
-    in
+  let pp_report_status ppf { unpack; commands; } =
+    let sep ppf () = Fmt.pf ppf ";@ " in
```
samoht: you can inline that `sep` value using `Fmt.unit`

src/git/negociator.ml
dinosaure: In the current implementation of Git, when you start a negotiation with a server, you send some hashes to notify to the server than you have these objects. The server will notice you the same. But we force a limit than if you send 128 hashes (see `_VAIN`), the client will abort the negotiation (with the flush packet or `0004done`).

From the git documentation (which one limits to 256 hashes) for the Smart protocol:
> (in the canonical implementation,
this is determined when the client sends 256 'have' lines without getting
any of them ACKed by the server - meaning there is nothing in common and
the server should just send all of its objects)

And for the HTTP Smart protocol:
> If the client has sent 256 "have" commits and has not yet
received one of those back from `s_common`, or the client has
emptied `c_pending` it SHOULD include a "done" command to let
the server know it won't proceed:
>
>   C: 0009done

So, `vain` is a value to count how many hashes (_vainly_ hashes) we sent and no one of these hashes was found in the server side (that means, the server never responds an ACK information).

dinosaure: I can change the value of `_VAIN` (by functor), I put 128 to test some complex computation between a client and a server.

src/git/smart.mli
dinosaure: The Smart decoder (and the encoder too) needs to know the length of the current hash implementation and the `Digest` module notify this information with `Digest.length`. I can reduce the signature need by these functors (like: `sig val length : int end`) if you want.

dinosaure:
From the commit 795c11b, we can switch between a `Bigarray` implementation or a `Bytes` implementation of the hash. I linted constraints when the module expected as the argument of the functor needs to provide:

```ocaml
module Digest =
struct
  type buffer = Cstruct.t

  val feed : ctx -> buffer -> unit

  ...
end

type hex = string

val of_hex : hex -> t
val to_hex : t -> hex

val of_string : string -> t
val to_string : t -> string

...
```

dinosaure:
I deleted the constraint about the error of the `FileSystem`, I keeped the polymorphic variant but the user is free to define a more precise error. I did this change because after, this could break the API (preferrably to do this change now).

dinosaure:
I just implemented the atomic computation for the reference (as @samoht requested as a deep requirement of `irmin`). I followed the previous implementation in the details, reviewed the code (about `EINTR` and other unix-dependent stuff), and matched to Mirage back-end (currently in the pipe).

I protected all mutable computation about the reference but I need to explain why I put an (other) _functor_ `Lock`. As I said, by the design, I concentrated all computation to be non-blocking and to only use some fixed-size buffers to improve the memory consumption (in general) of `ocaml-git`.

In this way, to read or write something, all of these are done in the __non__-atomic way. That means, in a concurrency context, we can write partially a Git object, switch to another process, which one read partially an other (or the same) Git object. About `tree`, `commit`, `blob` and `tag`, it does not matter because these Git object are immutable (it's forbidden for any process to modify these Git object).

Unfortunately, it's not the same for the `Reference`, which one is mutable. And it's the purpose of the `Lock` module, to protect for the race condition any modification of any `Reference`. So, we have the choice:

* We can expect an _atomic_ `write` function (which take care to `open`, `write` all and `close`, and - most important - `lock` this computation) from the `FileSystem` module (this is the case from the previous version)
* We need to provide a `Lock` abstraction to the `Reference` module to protect for any race-condition the `Reference.write` function - which one do the `open`, n-`write` and the `close` _syscall_ (this is the current case of this PR).

I don't have any strong-opinion about this choice (IMHO, I prefer the first one which is already available by `atomic_write` but not exposed). Then, because we can have the choice to change something about the API, I let this question for you, reviewer 👀 !

dinosaure:
In my shower time, I asked to myself the existence of the `lock_dir` value available in the Git state. Indeed, this value is close to the implementation of the `Lock` module: we consider a lock as a file, which will be stored in the `lock_dir`.

In the unix system (`git-unix`), this is what happen. We create a `*.lock` file to protect an update of the reference (as I said) against race condition.

However, in the mirage system, it's not the case. `lock_dir` is a unused value and we prefer to use an `Hashtbl` to represent this _bucket_ of _locks_.

The link between my previous comment and this thing is about the choice between some assumption of how we implement the _lock_ logic (highlighted by the `lock_dir` value) or the full abstraction of the _lock_ logic with an abstract type of how we store these _locks_ (file-system, hashtbl ?) and the lock itself.

It's a good argument for the second proposition, but I'm still open to other opinion and this is the week-end now.

dinosaure:
About the re-implementation of the `mem` module:

* Deleting of the PACK I/O: I understood clearly the purpose of the `write_pack` for a memory back-end, and obviously a _file-system_ back-end. So, I will work to a better provided interface and a `compact` function. But I need to notice than to read a PACK file, we need a virtual representation of a file-system - the point is: the de-serialization of the PACK file will be not available in the memory back-end.

* The common part of `store.ml` and `mem.ml`: I tried to be close the old interface of `store.ml` in the previous version with the new version. However, I keep the old interface for the `mem.ml`. This the main purpose of the PR: think about a new API. The new `store.ml` is what I think about the new interface (with some dependents functions to a virtual representation of the file-system - like reading a PACK file). The current `mem.ml` is the old interface (which all functions can be found in `store.ml`).

So, the point is, I can reduce the new `store.ml` API to the same interface than `mem.ml` or we can think about a new common interface.

* Serialization of the Value: this is the highest problem about the memory consumption in the previous version of ocaml-git. As I said in the introduction, all serialization/deserialization only needs some fixed-size buffer to write to a _file_. However, in the memory backend, we need to write to a buffer (`Cstruct.t`) which can be grow as a file. So, the memory back-end need a new functor, a `Value.BUFFER` which implement this idea. Obviously, we can use the `Buffer` module available in the standard library (with some little tricks to match with the interface) or create an optimized implementation.

The point and it's the goal of this implementation is to locate the possibly memory problem in this module - because it's the only part of the code where a buffer can be grow automatically (and when it can be difficult the infer the memory consumption).

And it's enough for the last commit :) !

src/git/mem.ml
```diff
@@ -0,0 +1,273 @@
...
+  val mem_reference : t -> Reference.t -> bool Lwt.t
+  val read_reference : t -> Reference.t -> Hash.t option Lwt.t
+  val read_reference_exn : t -> Reference.t -> Hash.t Lwt.t
+  val write_head : t -> Reference.head_contents -> unit Lwt.t
+  val read_head : t -> Reference.head_contents option Lwt.t
+  val write_reference : t -> Reference.t -> Hash.t -> unit Lwt.t
+  val remove_reference : t -> Reference.t -> unit Lwt.t
+  val test_and_set_reference : t -> Reference.t -> test:Hash.t option -> set:Hash.t option -> bool Lwt.t
+  val read_inflated : t -> Hash.t -> string option Lwt.t
+  val write_inflated : t -> Cstruct.t -> Hash.t Lwt.t
```
hannesm: not sure, but why is this within `Lwt.t`?  side effects I can observe below are mutation (of the hashtable) and errors (using either `raise` or `Lwt.fail`) -- I feel these functions could life outside of the Lwt monad...

dinosaure: Indeed, some parts of the `mem.ml` don't need to be wrapped in a `Lwt.t`. But we need to keep in our mind the compatibility between `store.ml` (which `Lwt.t` is needed - because we interact with the file-system) and `mem.ml` (and share the same API). So it's the global point (and this signature is come from the previous version of [ocaml-git](https://github.com/mirage/ocaml-git/blob/master/src/git/store.ml#L46)).

However, for this specific function, in the current version of the implementation of the `store.ml` with a file-system, I don't provide a `write_inflate` but only a [`write`](https://github.com/mirage/ocaml-git/blob/master/src/git/store.ml#L30) function.

The better question should be: Is it necessary to expose this function ? And, if the response is yes, how we can expose this kind of computation ? In fact, in my implementation, the signature should be:

```ocaml
val write_inflated_s : t -> Cstruct.t -> (Hash.t * int, error) result Lwt.t
```

`_s` notices than we use the _state-defined_ buffer (instead a `write_inflate_p : t -> Cstruct.t -> ztmp:Cstruct.t -> raw:Cstruct.t -> Cstruct.t -> (Hash.t * int, error) result Lwt.t`), the `Cstruct.t` is the already-serialized Git object, and we will return the `Hash.t` produced and how many byte(s) we wrote.

Finally, I think it's not needed (for the internal purpose of `ocaml-git`) to have this function (but a `read_inflated` function is needed in several points of the PACK encoder - so necessary in both-side, `store.ml` and `mem.ml`). Then, may be some peoples use `ocaml-git` and this function. We need to take care about that.

src/git/jbuild
```diff
@@ -1,7 +0,0 @@
-(jbuild_version 1)
```
dinosaure: I will switch to `jbuilder` (and learn in the same time how to use it). I use `ocamlbuild` just because it's more convenience for me 😄 !

dinosaure:
About the `Object_graph` module, I changed the `pack` function which one cannot use the `Lwt_list.map_p` to compute all Git object. Instead, we can use `Lwt_pool`, allocate needed buffer and compute some Git objects in the same time.

dinosaure:
Now, the current version of `ocaml-git` does not have __any__ dependency with C code. I replaced the C code to calculate the Rabin's fingerprint needed for the _delta-ification_ by an OCaml code. However, we have a strong limitation about the 32-bits architecture.

Indeed, the current implementation of the Rabin's fingerprint is close to the `unsigned int32` representation. I used  _hack_ of @dbuenzli (to `compare` two `unsigned int32`) from [mtime](https://github.com/dbuenzli/mtime/blob/master/src/mtime.ml#L27). However, when I want to cast the `unsigned int32` to a native integer, for a 32-bits architecture, it could possible to not be able to store the value in a native integer - I raise a `Failure` in this case.

But, this case could not happen because this case appear only when when we want to access to a value of some specific arrays which have some other assertions to avoid this possible case.

samoht:
Thanks for all the updates? Is the PR in a state when I can review it again?

Regarding locks: I agree that the second solution seems more elegant. About the in-memory implementation: it is pretty useful for testing purposes. I fully agree with modifying the API to be non-blocking for values, and add atomic/locking operations for references only. I think we should also keep some convenience functions to read/write full values.

Regarding pack files an in-memory stores: I don't understand what is the current limitation: are you able to read object in the pack but only write loose objects? If yes, do you have a way to re-compact  the pack files using the in-memory backend? I agree that this is a bit of a stretch use-case (as it doesn't matter to control the memory consumption as everything will stay in memory anyway) but it would be nice if it could somehow work :-)

dinosaure:
At this time, I implemented a small [common API](https://github.com/dinosaure/ocaml-git/blob/3b65ebc6bf1add433b8f2aec91358d15764141ff/src/git/minimal.mli) between the memory back-end and the store (file-system related back-end).

It's the time to think if it's the best and compare to the [store.mli](https://github.com/dinosaure/ocaml-git/blob/3b65ebc6bf1add433b8f2aec91358d15764141ff/src/git/store.mli) interface (which has more functions). Then, this interface will use to make the `Object_graph` and the `Search` module first, then the `Smart` and `Smart_http` protocol (with the `Negociation` engine and the `Revision` DSL).

samoht:
Thanks! 

The minimal API looks great. About the size: would it be possible to get the inflated size instead? Usually you want to pre-allocate some structure before reading some contents: in that case I am not sure what to do the the deflated size.

The store API looks ok. I am not sure I will need to use all these functions, but I guess it is nice we have them just in case :-) How do you feed data to the repository/pack file state? Is the internal decoder is similar to jsnonm/xmlm (with an `Await` state?)

Feel free to change the `Object_graph` and `Search` API as much as you want, I think they are just there to allow easy testing of the library. All the complex search are the Git DAG should be handled by Irmin anyway.

src/git/store.ml
```diff
@@ -194,17 +194,32 @@ sig
...
     val graph_p : t -> dtmp:Cstruct.t -> raw:Cstruct.t -> (Hash.t Reference.Map.t, error) result Lwt.t
     val graph : t -> (Hash.t Reference.Map.t, error) result Lwt.t
     val normalize : Hash.t Reference.Map.t -> Reference.head_contents -> (Hash.t, error) result Lwt.t
-    val list_p : t -> dtmp:Cstruct.t -> raw:Cstruct.t -> ((Reference.t * Hash.t) list, error) result Lwt.t
-    val list_s : t -> ((Reference.t * Hash.t) list, error) result Lwt.t
-    val list : t -> ((Reference.t * Hash.t) list, error) result Lwt.t
+    val list_p : t -> dtmp:Cstruct.t -> raw:Cstruct.t -> (Reference.t * Hash.t) list Lwt.t
+    val list_s : t -> (Reference.t * Hash.t) list Lwt.t
+    val list : t -> (Reference.t * Hash.t) list Lwt.t
+    val remove_p : t -> dtmp:Cstruct.t -> raw:Cstruct.t -> ?locks:Lock.t -> Reference.t -> (unit, error) result Lwt.t
```
samoht: why do you need intermediate buffers to remove a reference?

dinosaure: In this case, I check if the reference is `packed` or not. So, I need to decode the `packed-refs` file. One other solution is to _systematically_ `unpack` references when we want to create a Git state - and I think, it could be better about the performance because, at this stage, I need to check all the time where is located the reference (and decode for each operation the `packed-refs` file).

src/utils/lwt_lock.ml
```diff
@@ -0,0 +1,242 @@
+module Stack =
```
samoht: where does this code come from? please add copyright/license :-)

src/git-dot/ocaml_deflate.ml
```diff
@@ -1 +1 @@
-../git-unix/ocaml_deflate.ml
```
samoht: I don't undertand this commit. Why did you put everything under `utils`? Some of the files need to be linked with `lwt.unix`, others with `compiler-libs`, etc.

src/git/minimal.ml
```diff
@@ -134,4 +134,12 @@ sig
         from the git repository [state]. The [?lock] avoids the race
         condition if it's specified. *)
   end
+
+  val reset : ?locks:Lock.t -> t -> (unit, Ref.error) result Lwt.t
+  (** [reset ?locks t] removes all things of the git repository [t]
+      and ensures it will be empty. *)
+
+  val clear : ?locks:Lock.t -> t -> unit Lwt.t
```
samoht: `clear_caches` would probably be more explicit.

src/git-dot/ocaml_deflate.ml
dinosaure: I put all implementation in one directory and use `ln -sfn ...` in `git-unix`, `git-dot`, etc. to use these implementations to avoid any redundant code. For example, `fs_lwt_unix.ml` need to be available in the `git-dot` folder, `git-unix` and `git-top`, so we share a common implementation of the file-system available in one directory.

It's same for `sha1.ml` (need to be available in any sub-project).

The `utils` directory is not a project as `git-unix`, it does not contain any `jbuild` file and does not produce any libraries or binaries, it's just a bulk of implementations needed to make Git's functors.

IMO, it's better to do this for maintainability and avoid to apply a fix/change a multiple times for a same implementation. If we found a bug in `fs_lwt_unix.ml`, we fix it and this fix is directly available for other sub-project which use this specific implementation.

samoht: Yikes. Please move the core/common files into the "core" library (e.g. `git`) as it used to be and use it directly -- you should never need to either symlink or copy/duplicate files.

The `fs_lwt_unix.ml` is used by the unix backend only move it into the `git-unix` library. etc.

dinosaure:
About the issue #231 of the `?root` argument, we use `FileSystem.Dir.current ()` (same as `getcwd`) if the user does not notice the argument.

dinosaure:
From https://kate.io/blog/making-your-own-exploding-git-repos/ (repo available to [Katee/git-bomb](https://github.com/Katee/git-bomb)). `ocaml-git` can `clone` (with `ogit-http-clone` - I fixed some details about the URI) and produce a `*.dot` (with `git_dot.exe`) without any explosion of your memory 👍 !

src/git-unix/ogit-http-fetch/main.ml
```diff
@@ -0,0 +1,293 @@
...
+    let body = match body with
+      | None -> None
+      | Some stream ->
+        Lwt_stream.from stream
+        |> Lwt_stream.map (fun (buf, off, len) -> Cstruct.to_string (Cstruct.sub buf off len))
+        |> fun stream -> Some (`Stream stream)
+    in
+
+    (* XXX(dinosaure): [~chunked:true] is mandatory, I don't want to
+       explain why (I lost one day to find this bug) but believe me. *)
```
samoht: you use `chunked:false` in the code and `chunked:true` in the comment :-)

dinosaure:
At this stage, `test_common` is done.

samoht:
Could you rebase your branch on master? I'll make a final review and then we can merge.

samoht:
The tests on 4.04 seems to pass now (yay!). The errors on 4.03 should be fixed by https://github.com/ocaml/opam-repository/pull/10897

samoht:
So now all the tests pass on Travis, the failures are due to missing constraints in the revdeps. These should be fixed by https://github.com/ocaml/opam-repository/pull/10904 (this is not a blocker for that merge).

The tests are failing on windows for git-mirage with:
```
- [exec] cmd /d /v:off /c rd /s /q "test-git-mirage-store"
- The system cannot find the file specified.
- [exec] error 2
```
which I guess is just missing a `Sys.file_exists` somewhere in the test init code.

Also the unix tests are failing on windows:
```
[ERROR]             Unix Store             4   Operations on references.
...
#=== ERROR while compiling git-unix.dev =======================================#
# opam-version         1.3.0~dev (2c2399d98ad0ebcb43a4bc856f3636030bb376db)
# os                   win32
# command              jbuilder runtest -p git-unix
# path                 C:/cygwin64/home/appveyor/.opam/4.03.0+mingw64c/build/git-unix.dev
# exit-code            1
# env-file             C:/cygwin64/home/appveyor/.opam/4.03.0+mingw64c/build/git-unix.dev\git-unix-288-d2c37b.env
# stdout-file          C:/cygwin64/home/appveyor/.opam/4.03.0+mingw64c/build/git-unix.dev\git-unix-288-d2c37b.out
# stderr-file          C:/cygwin64/home/appveyor/.opam/4.03.0+mingw64c/build/git-unix.dev\git-unix-288-d2c37b.err
### stderr ###
# [...]
# ASSERT r1
# --------------------------------------------------------------------------------
# --------------------------------------------------------------------------------
# ASSERT r2
# --------------------------------------------------------------------------------
# [exception] "Assert_failure src/git/reference.ml:309:8"
# Raised at file "bytes.ml", line 219, characters 25-34
# Called from file "src0/sexp.ml", line 90, characters 13-47
# 
# 
# The full test results are available in `_build/_tests`.
# 1 error! in 0.094s. 16 tests run.
```

samoht:
More revdeps fixes there: https://github.com/ocaml/opam-repository/pull/10910

samoht:
The current status is:
- all the test runs are green on Linux / Travis CI
- the test runs are green on Appveyor / Windows for git-unix but they fail for an unknown reason on git-mirage. It is very hard to investigate the windows error without a windows environment that neither @dinosaure or I have at the moment so let's say this is not a blocker for the merge but should be fixed ASAP.

So I am happy to merge that PR in this current state. Thank you very much @dinosaure for all that hard work. The new `ocaml-git` (that I will probably bump to 2.0) is a great piece of code :-)

PR #229, opened 2017-09-22
Set default path according to documentation

cuihtlauac:
Set default path to "." instead of "/tmp/ogit".

### DIFF
--- a/src/git/fs.ml
+++ b/src/git/fs.ml
@@ -144,7 +144,7 @@ module Make (IO: IO) (D: Hash.DIGEST) (I: Inflate.S) = struct
   let temp_dir t = t.dot_git / "tmp"
   let lock_dir t = t.dot_git / "lock"
 
-  let create ?(root="/tmp/ogit") ?dot_git ?(level=6) () =
+  let create ?(root=".") ?dot_git ?(level=6) () =
     if level < 0 || level > 9 then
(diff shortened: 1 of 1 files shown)


jnavila:
Doh! That simple?

samoht:
I am a bit reluctant for this change, as I already wiped the repository I was working in at least twice with that default (it was like this initially). Maybe it is safer to change the documentation and to force that parameter to be required (e.g. remove the `?`). WDYT?

jnavila:
I'm fine with forcing the user to clearly specify the repo in which the program is working. Definitely, the option's default would be the current working directory, but as you mentioned, that could surprise the developer. I had some hard time finding where the regression was coming from, the error message in case there's no .git directory being very confusing.

cuihtlauac:
I feel the pain on accidental wipes. I'm only reading .git carrying repositories. To me, having "/tmp/ogit" as default hardwired is a clumsy defence. A parameter without default value sounds like a cleaner solution.

samoht:
As there is a big refactoring in progress (#227) I won't make that API change yet -- but I'll keep that change in mind. Until them, I will just fix the README (see #210). Thanks for your feedback!

samoht:
See #231 for the API change. I am closing this issue as I have now updated the README with the correct runes.

Issue #223, opened 2017-07-04
The example in the README file no longer works


jnavila:
```
# Lwt_main.run (read ["README.md"] >|= print_string)
;;
Exception: Failure "Git.FS.read_reference_exn: refs/heads/master not found".
Raised at file "src/core/lwt.ml", line 805, characters 16-23
Called from file "src/unix/lwt_main.ml", line 34, characters 8-18
Called from file "toplevel/toploop.ml", line 180, characters 17-56 
```

cuihtlauac:
PR Fix: https://github.com/mirage/ocaml-git/pull/229

PR #230, opened 2017-09-26
Update README

samoht:
See #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Performance is comparable to the Git tool.
 # module Search = Git.Search.Make(FS);;
 
 # let read file =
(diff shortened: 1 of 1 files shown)


samoht:
The revdep failure is unrelated:

```
# File "_none_", line 1:
# Error: No implementations provided for the following modules:
#          Str referenced from /home/travis/.opam/4.03.0/lib/prometheus-app/prometheus-app.cmxa(Prometheus_app)
```

@talex5 any idea?

talex5:
For some reason, it's using a very old version of prometheus-app with the latest version of prometheus. I've made a PR to fix that here: https://github.com/ocaml/opam-repository/pull/10357

PR #236, opened 2017-12-05
Fixed the fetch command

dinosaure:
It's not properly a fix but an improvement of the interface of the HTTP sync module:
* instead to use `string` as a reference, we use `Reference.t`
* we check if the hash of the client is the same as the server (for _exactly_ the same reference)
* we provide the function `fetch_some` which fetches some references (not all, not one)

This PR is not done yet, we need to move this modification to the tcp sync protocol and factorize the code - `fetch_some` is more abstract than `fetch_one` or `fetch_all`.


dinosaure:
We need to talk about the API on the `fetch` command. As I said, this PR does not fix a big bug - it's about a more about a complex API on a complex command.

The `fetch` command is complex and `Sync.fetch` do absolutely nothing, like:
* it does not write on any references - so we downloads objects, then it returns all _server-side_ references with the new hash
* it does not care about if it's needed to download or not a commit - I mean, it does not check if you have already the hash commit from the server on your side

This function just `fetch` and let the user update references and take care about the `want` function - and checks if it's necessary to download or not a remote reference.

About the bug catched by @AltGr, it's more about the complex API than the bug - however he catches a bug about the negotiation engine - when he does not know what he needs to check before and what he need to do after.

`fetch_{some,all,one}` could be a good start to provide a better function juste to fetch without a strong acknowledge of Git. Currently, the API stills complex:

* `fetch_all` has 2 importants arguments:
  * `references`, it's an associated list between local reference and remote reference. This list can be not exhaustive - but we continue to fetch all references. Then, for any updated remote reference, we will write the associated local reference with the new hash.
  * `origin`. However, we still fetch all references. For remote reference with a _common_ pattern (like `refs/heads/{branch}`), we will store hashes of these reference to `refs/remotes/{origin}/{branch}`. For any other remote reference, we return an associated list between these remote references and the hash associated

The first list returned is a list which contains all updated local reference, the second list, see below.

* `fetch_some` is the more specific function which fetch only what is specified by the client - in this case, we should not discard any reference from the server and just returner all updated local references

* `fetch_one` is a specific case of `fetch_some`

For all of these function, we have a `want` _engine_. This algorithm reduce or not the list of commits to fetch from remote references. Then, it checks if these commits already exists on the local store and, if it's the case, discards this commit to fetch.

For me, `fetch_all` stills too complex and I prefer to not save automatically some references because they respect an arbitrary pattern - and prefer to move on these remote references as discard updated references on the second list.

But if you have a better API, I'm listening. The current situation is not good - @AltGr lost lot of time about that.

samoht:
@dinosaure is this good to merge?

dinosaure:
Not yet :(

dinosaure:
Now the API seems to be good but not ready to merge, @AltGr seems to catch a new bug.

PR #238, opened 2017-12-07
Add infix operator to make a reference

dinosaure:
Convenience PR

### DIFF
--- a/src/git/reference.mli
+++ b/src/git/reference.mli
@@ -21,7 +21,7 @@ type t = private string
 (** A Git Reference object. Which contains a hash to point to an other
     object. *)
 
-module Infix : sig
+module Path : sig
   type partial
   (** Type of the left part of a reference. *)
   type branch = string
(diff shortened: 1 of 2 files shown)


dinosaure:
This PR add some useful _infix_ operator to make easily a `reference`.

src/git/reference.mli
```diff
@@ -21,6 +21,31 @@ type t = private string
...
+  type branch = string
+  (** Tyoe of the right part of a reference. *)
+
+  val ( // ): partial -> string -> partial
+  (** Infix operator to compose [string] value with a {!partial} reference. *)
+
+  val ( / ): partial -> branch -> t
+  (** Infix operator to compose {!partial} with {!branch} and return a full-defined reference {!t}. *)
+
+  val refs: partial
```
samoht: this is not really `Infix`, right? Maybe just move all the code to the top-level so that a user coud do:

```
Git.Reference.(heads / master)
```

samoht: or name it `Path` instead of `Infix` (to not conflict with other top-level definitions)

dinosaure:
Ready!

PR #239, opened 2017-12-07
Fix serialisation of references when using test and set

samoht:


samoht:
I am not sure this is the right fix as it is strange that we use a different code path than for `set`. Also the other `XXX` comment in there is not very re-assuring. 

If we write without an ending `\n`, the decoder will just loop ever. We clearly need more tests!

dinosaure:
This the right fix, the decoder [expect a \n at the end of the reference](https://github.com/mirage/ocaml-git/blob/master/src/git/reference.ml#L186). I fixed the behaviour from the encoder before - but don't catch this computation then.

And, about the question on why we don't use the encoder on the `test_and_set` function, is a good question. May be, the best is to use, as you said, the encoder instead `Fmt.pf`.

samoht:
Note: a similar issue (but with a different fix) is https://github.com/moby/datakit/pull/611

As the Git client always put a '\n' I think it is enough to make `ocaml-git` always write a newline as well. I have tried a bit to use the same code path as `write` but I didn't find how to serialise a value to a string ... so I'll merge as-is but please @dinosaure feel free to improve that :-)

PR #241, opened 2017-12-12
Improve the convenience function to fetch and push command for the TCP back-end

dinosaure:
This is the same way from the fix-fetch branch which improve the API about fetch & push command, we could update tests with this new API.


src/git-http/sync.ml
```diff
@@ -454,23 +417,28 @@ module Make_ext
...
-          List.map (function
-              | `Create (hash, reference) -> Encoder.Create (hash, Store.Reference.to_string reference)
-              | `Delete (hash, reference) -> Encoder.Delete (hash, Store.Reference.to_string reference)
-              | `Update (_of, _to, reference) -> Encoder.Update (_of, _to, Store.Reference.to_string reference))
-            commands
-          |> fun commands -> List.hd commands, List.tl commands
-        in
-
-        packer ~window:(`Object 10) ~depth:50 ~ofs_delta:true git refs.Decoder.refs (x :: r) >>= function
+        Lwt_list.map_p (function
```
samoht: why do you use `map_p` here? There is no blocking operation so everything will happen sequentially anyway.

test/git-mirage/test.ml
```diff
@@ -133,9 +133,28 @@ module TCP = Test_sync.Make(struct
...
+        Lwt.try_bind
+          (fun () -> Lwt_list.iter_s
+              (fun (remote_ref, new_hash) ->
+                 Store.Ref.write t remote_ref (Store.Reference.Hash new_hash) >>= function
+                 | Ok () -> Lwt.return ()
+                 | Error err -> Lwt.fail (Jump err))
+              (Store.Reference.Map.bindings downloaded))
+          (fun () -> Lwt.return (Ok ()))
+          (function Jump err -> Lwt. return (Error (`Ref err))
+                  | err -> Lwt.fail err)
```
samoht: all of this will probably be copy/pasted by users of the TCP sync. So this should probably go into the `Sync` module ...

src/git/sync_common.ml
```diff
@@ -0,0 +1,147 @@
+module Make
```
samoht: can you move this into the `Sync` module? Not sure we need/want to create a new module name for this (people will wonder what's the difference between `Sync` and `Sync_common`)

test/git-mirage/test.ml
dinosaure: About this part of the PR is a side effect of the new API in 2 terms (the bottom and the top).

At the bottom, it's only about API.

For me, `fetch_all` should not automatically create the references itself. It creates references only according to client wishes (by the arguments `~references`).

However, `fetch_all` downloads all references and if the references argument is not exhaustive (and, I believe it's a common case), the client could lost some useful hashes endpoints (because they are not saved in references).

Before this patch, I provided this signature:

```ocaml
val fetch_all:
   Store.t -> ?locks:Store.Lock.t ->
   ?capabilities:Capability.t list ->
   create:(Store.Reference.t -> Store.Reference.t option Lwt.t) ->
   references:Store.Reference.t list Store.Reference.Map.t ->
   Uri.t -> (Store.Hash.t Store.Reference.Map.t
             * Store.Reference.t list Store.Reference.Map.t
             * Store.Hash.t Store.Reference.Map.t, error) result Lwt.t
```

As you can see, we expect a `create` function which is a process to make a new local reference from the already downloaded remote reference. Then, we save the hash endpoint to this new local reference (which does not appear in references).

I'm not very happy about this previous signature for two reasons:
- firstly, we complexify a lot the API - we need to explain to the client: "ok, all references in the references will surely be created but the create function will create local references too"
- secondly, I don't like to let the user to define a black box function

And because I don't like to let `fetch_all` to create arbitrary local references, I prefer to have this current API.

At the top, it's only about what the test expect.

When I did this patch, I thinked to keep the previous same semantic about `fetch_all` and I did not think if it's necessary to have systematically all local references binded with remote references - I did not look where this function was used and if it's necessary to have this semantic.

If I remember correctly, the test on the `fetch_all` works only on one reference, master. It could be (may be?) more useful to provides `~references` as an argument of the `fetch_all` function available on the [SYNC](https://github.com/mirage/ocaml-git/blob/master/test/common/test_sync.ml#L21) signature and explicitly want to save the remote master on the local master.

dinosaure:
Just to move my old comment to the top-level about this question.

> all of this will probably be copy/pasted by users of the TCP sync. So this should probably go into the Sync module ...

About this part of the PR is a side effect of the new API in 2 terms (the bottom and the top).

At the bottom, it's only about API.

For me, `fetch_all` should not automatically create the references itself. It creates references only according to client wishes (by the arguments `~references`).

However, `fetch_all` downloads all references and if the references argument is not exhaustive (and, I believe it's a common case), the client could lost some useful hashes endpoints (because they are not saved in references).

Before this patch, I provided this signature:

```ocaml
val fetch_all:
   Store.t -> ?locks:Store.Lock.t ->
   ?capabilities:Capability.t list ->
   create:(Store.Reference.t -> Store.Reference.t option Lwt.t) ->
   references:Store.Reference.t list Store.Reference.Map.t ->
   Uri.t -> (Store.Hash.t Store.Reference.Map.t
             * Store.Reference.t list Store.Reference.Map.t
             * Store.Hash.t Store.Reference.Map.t, error) result Lwt.t
```

As you can see, we expect a `create` function which is a process to make a new local reference from the already downloaded remote reference. Then, we save the hash endpoint to this new local reference (which does not appear in references).

I'm not very happy about this previous signature for two reasons:
- firstly, we complexify a lot the API - we need to explain to the client: "ok, all references in the references will surely be created but the create function will create local references too"
- secondly, I don't like to let the user to define a black box function

And because I don't like to let `fetch_all` to create arbitrary local references, I prefer to have this current API.

At the top, it's only about what the test expect.

When I did this patch, I thinked to keep the previous same semantic about `fetch_all` and I did not think if it's necessary to have systematically all local references binded with remote references - I did not look where this function was used and if it's necessary to have this semantic.

If I remember correctly, the test on the `fetch_all` works only on one reference, master. It could be (may be?) more useful to provides `~references` as an argument of the `fetch_all` function available on the [SYNC](https://github.com/mirage/ocaml-git/blob/master/test/common/test_sync.ml#L21) signature and explicitly want to save the remote master on the local master.

src/git-http/sync.ml
```diff
@@ -417,13 +417,13 @@ module Make_ext
             (fun _ -> Lwt.return ())
         in
 
-        Lwt_list.map_p (function
+        Lwt_list.map_s (function
```
samoht: Actually I was more thinking of `List.map` here :-)

dinosaure:
Ready to merge.

dinosaure:
About this PR. because I change a little bit the behaviour of the Smart protocol (used by the HTTP overlay and the TCP overlay), I checked if the behaviour does not change on the TCP way (because I well tested this path only on the HTTP overlay).

So all seems to be good and ready to merge. I did not change anything because I don't want to move the logic from the HTTP overlay (when we check if the negotiation engine return an empty list or not) on the Smart protocol.

src/git-http/sync.ml
```diff
@@ -736,7 +736,47 @@ module Make_ext
...
+    in
+
+    Lwt_list.filter_map_s
+      (fun action -> match action with
+        | `Update (remote_hash, local_hash, _) ->
+          Store.mem git remote_hash >>= fun has_remote_hash ->
+          Store.mem git local_hash >>= fun has_local_hash ->
+
+          if has_remote_hash && has_local_hash
+          then Lwt.return (Some action)
```
samoht: you can use `>|=` and remove the `Lwt.return` statements in both branches.

src/git-http/sync.ml
```diff
@@ -736,7 +736,47 @@ module Make_ext
...
+          else Lwt.return None
+        | `Create (local_hash, _) ->
+          Store.mem git local_hash >>= function
+          | true -> Lwt.return (Some action)
```
samoht: same here.

PR #243, opened 2017-12-14
Fix git-mirage on windows

samoht:


samoht:
So it seems that I fixed `git-mirage` but broke `git-unix` on windows... ho well...

samoht:
> If I understand correctly, most of errors about fpath come from un-normalized path (which has some empty segments). May be, instead to have a fapth_to_string function, it's better to use Fpath.to_string (Fpath.normalize path)).

One the issue is the handling of `./` which doesn't seem to work fine on some windows environnements. Unfortunately `Fpath.(normalize @@ v "./") = "./"` so this won't help much here.

For the `test_and_set` functions I plan to revisit that later on, as it's probably something that `irmin-git` should do itself.

samoht:
But the good news is that everything seems green now!

PR #244, opened 2017-12-15
Fixed fetch command over the HTTP protocol.

dinosaure:
All is explained in the src/git-http/sync.ml comment to understand
what happen and how to fix it. It's much more about an unspecified
behaviour than a bug.

I will explain later about this PR, I need to sleep.

### DIFF
--- a/src/git-http/sync.ml
+++ b/src/git-http/sync.ml
@@ -572,11 +572,7 @@ module Make_ext
              |> (fun uri -> Uri.with_port uri port))
         in
 
-        negociation_request `Flush has >>= fun resp ->
-
-        Log.debug (fun l -> l ~header:"fetch" "Receiving the first negotiation response.");
-
-        let result resp =
+        let negociation_result resp =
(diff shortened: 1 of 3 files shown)


dinosaure:
About this PR, we change the semantic of `Smart.Decoder.p_negociation` which now stop when the noticed `has` argument was entirely found in the server response.

Then, we can check the negotiation result.

This change is because, we can not differentiate (without alteration) the negotiation response which starts with ACK for each packets and the negotiation result which can start with ACK.

So, if the `has` was entirely found in the response, that means we does not expect other negotiation line.

dinosaure:
Not to merge, I really need to figure about the negotiation protocol and the diff behaviour about TCP and HTTP protocol. May be it's not the best solution.

dinosaure:
Ready to merge - If you want to test this PR, you need to merge #261 before (some commits in this repository use extra headers and you will fails otherwise).

PR #247, opened 2017-12-23
Index file fully implemented

dinosaure:
This PR show the current implementation of the serialization/deserialization with a `Container` structure which is the representation of the `index` file in OCaml. This implementation *is not* well tested but could be a good entry point to test all `ocaml-git`.

Indeed, `git` use `write-tree` and `update-index` (after it checks expected behaviors about some UNIX command on the file-system) to check basics operations on a git repository.

This module has, however, two problem currently:
* it depends on `Lwt_unix` (`stat` _syscall_) - so with this module, we put `Lwt_unix` and `Unix` as a strong dependency of `git` - it's already the case by the `mmap` syscall, however, it seems than OCaml maintainers want to delete this dependency by this way, so if it's done, `ocaml-git` will be continue to need to be linked with `Unix` (and could be a problem for a MirageOS environment).
* Currently, we serialize the version 2 of the index file when the version 4 is available. I can not test if my serialization on the version 4 works because my `git` (from the official repository) stills to use the version 2. I need to figure about that.

And, as I said, this module was just tested in an easy case and need some tests to be more robust.


dinosaure:
To show how to use the new INDEX module, I implemented the basic command `git write-tree` which walk on the INDEX file, then it makes new object _tree_ from the current _blobs_ available. This command line could be a good one to start more generally test case.

Indeed, `git` starts to test itself with this command and check if the returned hash of `write-tree` is as expected (it try with a already know file and an empty repository).

Next goal is to implement `update-index` and follow the git test suite.

dinosaure:
Close by #285

PR #248, opened 2017-12-28
Improve test

dinosaure:
This PR improved the test from an idea talked with @samoht. One of the big problem about testing is we don't have enough data to test all. The idea is to get some _mini_ PACK file (IDX and packed-refs - the last one is not yet implemented) to import them and have good complete tests cases and avoid to write them with our hands (like `test_store.ml`).

So, I created a new module `test_data.ml` which import and check PACK file and compare with the git output.

This code factorize a lot the test code and it's a good start for some others tests like specifically (and the final goal) `git rev-list`.


test/common/test_data.ml
```diff
@@ -0,0 +1,187 @@
...
+          Git.Buffer.add buf (Cstruct.sub ctmp 0 (E.used_out t));
+          Store.Hash.Digest.feed ctx (Cstruct.sub ctmp 0 (E.used_out t));
+        end
+    in
+
+    let () = go (Store.Hash.Digest.init ()) (E.default (R.to_sequence tree) hash_pack) in
+
+    if String.equal (load_file Source.idx) (Git.Buffer.contents buf)
+    then Lwt.return (Ok ())
+    else Alcotest.fail
```
samoht: you can use `Alcotest.failf`

dinosaure:
Can not reproduce the fail on `git-mirage` with same packages (eg. `mirage-fs-unix.1.4.1`) 😞 ...

dinosaure:
Revived.

dinosaure:
jbuilder, travis is not happy ...

jpdeplaix:
Looks like it's the same problem as in https://github.com/ocaml/dune/issues/483

dinosaure:
I don't understand why we have an error on the `git-mirage` back-end.

dinosaure:
So, about Travis, we fail on `REVDEP` and about appveyor, we fail when we compile `camlzip` for `git-unix` and `mirage-fs-unix` for `git-mirage`.

samoht:
The revdep error is fixed by https://github.com/ocaml/opam-repository/pull/11459
The appveyor issue is probably related to the recent switch to `jsonm`: it seems that https://ci.appveyor.com/project/samoht/ocaml-git-7vmk0/build/1.0.532/job/v5vaqv10b4rh62o1#L779 is parsed as an empty list of packages /cc @jpdeplaix

samoht:
I've relaunched the failing runs now that ocaml/opam-repository#11459 and https://github.com/ocaml/ocaml-ci-scripts/pull/220 are merged. Let's see how it goes (hopefully green)

samoht:
Last error:

```
$ ./test.exe test -v "mem: usual" 0
```

is working fine unders Linux/MacOS but is blocking forever on windows.

samoht:
I fixed the `verify_pack` tests on windows, by using `open_in_bin` instead of `open_in`. There is an other error related to unpack now, investigating ...

dinosaure:
👍

PR #252, opened 2018-01-04
Add an iter function on tree (to avoid cast with list)

dinosaure:


dinosaure:
Not to merge, see #254 which integrates this patch.

samoht:
Merged in #254

PR #253, opened 2018-01-04
Locals improvements

dinosaure:
This patch is just a little improvement in some parts of the code (but necessary to merge).


dinosaure:
Not to merge, see #254 which integrates this patch. This patch fails because it needs #252.

PR #254, opened 2018-01-04
Bug to generate a PACK file

dinosaure:
This PR is a merge of #252 and #253 plus a fix about the PACK file.

Again it's about the ownership of the `Cstruct.t` send to the PACK algorithm. Then, this function take care to not delta-ify multiple times the same object. About the last change, may be it's more useful to expect from the client a `Set` of objects instead a list.


samoht:
That looks good to me.

PR #260, opened 2018-01-04
Fix push

dinosaure:
This code fixed 3 kinds of bugs:
- parsing of the Smart protocol on the HTTP overlay (this patch does not have conflict with the parsing on the TCP overlay)
- check if it's needed or not to update a the server's reference (`push_handler` algorithm) and avoid to send an empty PACK file
- better algorithm to calculate which objects we need to send to the server (an algorithm which color the DAG) and avoid to traverse entirely the repository - see [rev-list](https://github.com/dinosaure/rev-list)

We can start to test this implementation.

### DIFF
--- a/src/git-http/sync.ml
+++ b/src/git-http/sync.ml
@@ -461,12 +461,21 @@ module Make_ext
              |> (fun uri -> Uri.with_host uri (Some host))
              |> (fun uri -> Uri.with_port uri port))
           >>= fun resp ->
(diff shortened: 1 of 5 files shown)


dinosaure:
This PR was tested with #244 with [this code](https://gist.github.com/dinosaure/8acc87541918796df7086721101f4509) with [a well defined token](https://gist.github.com/dinosaure/8acc87541918796df7086721101f4509#file-test_fufpf-ml-L173) to push to the repository [test_fetch_some](https://github.com/dinosaure/test_fetch_some).

PR #262, opened 2018-01-09
WIP: remove the lock module and test_and_set

samoht:
This is work in progress to clean-up the concurrency invariants in the FS backend. In the current implementation, there are 2 functions which need to be atomic:
- move/rename (because POSIX)
- test_and_set (because Irmin)

It would be much easier to rely only on `move` and let Irmin does all the annoying locking to guarantee the `test_and_set` is atomic. This branch is doing that:

- it removes the `Lock` module, and all the optional `lock` arguments
- it removes `test_and_set` completely (which simplifies a lot of things)

The tests seems to pass still fine, but I am not surprised as only Irmin is testing concurrency stuff at the moment. My plan is to check if I can move all the locking in Irmin directly so that `ocaml-git` just focuses on the core Git logic, as it should always have been the case.


samoht:
I am starting to be happy with these changes. Lots of cleanups, and the resulting concurrency semantics is safer.

dinosaure:
The last commit seems to be very good for me 👍 !

samoht:
The last commit fixes a race in `File.move` the git-mirage backend. With this all the current tests pass reliably.

PR #263, opened 2018-01-10
Remove test_and_set operations

samoht:
This is not possible to ensure it without locking everything which is
slow and uncessary. At an higher level, Irmin could do it by enforcing
a single-writer policy, so let Irmin handle this and simplify ocaml-git
instead.


src/git-unix/misc.ml
```diff
@@ -32,26 +32,6 @@ let is_dir path =
...
-   Using read is necessary to handle references, since these are
-   mutable and can't be cached. Using mmap here leads to hitting the
-   OS limit on the number of open files.
-
-   This threshold must be larger than the size of a reference.
-
-   XXX(dinosaure): About the PACK file, we use directly mmap by the
-   [MAPPER] abstraction to compute efficiently all. I don't know if
-   it's reliable to keep this behaviour when we ensure than we use
-   [read] only for the loose object and mmap for the PACK file.
```
samoht: @dinosaure do you know if this comment is still up-to-date? What happen if one tries to open the same file a lot? Do we run out of inodes? Is seems that we don't have that weak cache of fd anymore. See https://github.com/mirage/ocaml-git/pull/143 for the initial PR.

dinosaure: To the current implementation the PACK was handle in 2 two:
1) when we meet a __new__ PACK file (without the associated IDX file), we read entirely the PACK file (with `read` syscall) continuously to produce the IDX file and calculate some stuff.
2) for all others, we use only the [MAPPER] abstraction (and, by this way, the `mmap` syscall) and keep file-descriptors of these PACK files along than the program run (so, one file-descriptor per PACK file - exactly one for the IDX file and one for the PACK file)

This is only about PACK file. And it's why, then, I explained it could be relevant to keep this for loose (big) object like blob. So we can delete the pool of `open` if you want when, only for PACK files, we use directly `mmap` - see [Unpack.Decoder](https://github.com/mirage/ocaml-git/blob/master/src/git/unpack.ml#L1394) and specifically [optimized_get](https://github.com/mirage/ocaml-git/blob/master/src/git/unpack.ml#L1964).

dinosaure: So, I believed this comments stills up-to-date :) .

samoht: But that `mmap_threashold` variable is only used when doing a test-and-set, e.g. for modifying references. I don't understand how this is used for large blobs.

dinosaure: Ah, yeah. You are true, so if you delete `test_and_set`, you should delete `mmap_ threshold` because ... And because the reference size should not upper than `mmap_threshold`, it's useless to keep it.

PR #266, opened 2018-01-15
Remove _p and _s functions

samoht:
Add a new `buffer` type which and add a few buffers to Git repositories' state.
Add a few convenience functions to allow sharing these buffers via an Lwt_pool.

The concurrent behavior of all these functions should now be clearer and safer.


dinosaure:
All is good for me 👍

PR #268, opened 2018-01-16
Weave in the file-system state in every filesystem calls

samoht:
This allows to remove the 'GAMMA' functor parameter and to cleanup the
`create/connect` functions. Now the main store signatures do not define a
creator anymore, only the implementation do and they can take different
parameters (e.g. the typical MirageOS style).


samoht:
Rebased.

samoht:
@dinosaure acked the changes offline, so merging.

PR #271, opened 2018-01-18
Clean binaries

dinosaure:

PR #272, opened 2018-01-18
Clean error

dinosaure:


src/git/commit.ml
```diff
@@ -33,7 +33,7 @@ sig
   module D: S.DECODER
     with type t = t
      and type init = Cstruct.t
-     and type error = [ `Decoder of string ]
+     and type error = Error.Decoder.t0
```
samoht: `t0` is not a very good error name. I understand the separation in "levels" but the result is not so great.

samoht:
After a quick I am globally happy with these changes, this simplifies quite a bit the way errors are handled. But I think we should even go further in the simplification process and make most of the error partially abstract and do not distinguish between all the `t0`, `t1`, `t2`(just use `t`). For every error, one needs to check if there is noone pattern matching on it, and if yes just make it abstract (the constructors should stay there as they are used to "raise" specific errors).

dinosaure:
I push 2 merge, one it's about #271 (which compiles) and the other is #270 (already merged) to fix conflicts and compilation.

Then, as @samoht said, I merged `t0`, `t1`, etc. to `t`. I use functor too instead to hand-write all errors in signatures.

samoht:
Sorry to nitpick but can you use a polymorphic type `'a error` instead of having yet an other functor?

samoht:
LGTM. Can you rebase it?

samoht:
Final pushes look very good to me.

I still really dislike the kind of code that we have in https://github.com/mirage/ocaml-git/pull/272/commits/c7a0d52c65c86b13892cb28e5d866768016d8970#diff-9b56c2745e0b5716be32fdb28526c4b1R129 where we have way too many operators. But this is out of scope of that PR :-)

PR #273, opened 2018-01-18
Avoid functor nightmare

dinosaure:
Need #272

### DIFF
--- a/src/git/pack.ml
+++ b/src/git/pack.ml
@@ -90,9 +90,9 @@ sig
   val topological_sort: t list -> t list
 end
 
-module Entry (H : S.HASH) =
+module MakeEntry (Hash: S.HASH): ENTRY with module Hash = Hash =
 struct
-  module Hash = H
+  module Hash = Hash
 
   type t =
     { hash_name   : int
@@ -161,6 +161,13 @@ struct
     ; length }
 
   let id { hash_object; _ } = hash_object
+  let kind { kind; _ } = kind
+  let preferred { preferred; _ } = preferred
+  let delta { delta; _ } = delta
+  let length { length; _ } = length
+
+  let with_delta t delta = { t with delta }
+  let with_preferred t preferred = { t with preferred }
 
   let name x name =
     { x with hash_name = hash name
--- a/src/git/pack.ml
+++ b/src/git/pack.ml
@@ -53,6 +53,43 @@ struct
     | Tag -> Fmt.pf ppf "Tag"
 end
 
+module type ENTRY =
+sig
+  module Hash: S.HASH
+
+  type t
+
+  type source =
+    | From of Hash.t
+    | None
+
+  val pp: t Fmt.t
+  val pp_source: source Fmt.t
+  val hash: string -> int
+
+  val make:
+    Hash.t
+    -> ?name:string
+    -> ?preferred:bool
+    -> ?delta:source
+    -> Kind.t
+    -> int64
+    -> t
+
+  val kind: t -> Kind.t
+  val preferred: t -> bool
+  val delta: t -> source
+  val length: t -> int64
+
+  val with_delta: t -> source -> t
+  val with_preferred: t -> bool -> t
+
+  val id: t -> Hash.t
+  val name: t -> string -> t
+  val compare: t -> t -> int
+  val topological_sort: t list -> t list
+end
+
 module Entry (H : S.HASH) =
 struct
   module Hash = H
@@ -218,6 +255,34 @@ struct
   let ( - )  = Int64.sub
   let ( >> ) = Int64.shift_right
   let ( / ) = Int64.div
+end
+
+module type H =
+sig
+  module Hash: S.HASH
+
+  type error
+
+  val pp_error: error Fmt.t
+
+  type t
+
+  type reference =
+    | Offset of int64
+    | Hash of Hash.t
+
+  val pp: t Fmt.t
+  val default: reference -> int -> int -> Rabin.t list -> t
+  val refill: int -> int -> t -> t
+  val flush: int -> int -> t -> t
+  val finish: t -> t
+  val eval: Cstruct.t -> Cstruct.t -> t ->
+    [ `Await of t
+    | `Flush of t
+    | `End of t
+    | `Error of (t * error) ]
+  val used_in: t -> int
+  val used_out: t -> int
 end
 
 module MakeHunkEncoder (H : S.HASH) =
@@ -539,6 +604,36 @@ struct
   let used_out t = t.o_pos
 end
 
+module type DELTA =
+sig
+  module Hash: S.HASH
+  module Entry: ENTRY with module Hash := Hash
+
+  type t =
+    { mutable delta: delta }
+  and delta =
+    | Z
+    | S of { length    : int
+           ; depth     : int
+           ; hunks     : Rabin.t list
+           ; src       : t
+           ; src_length: int64
+           ; src_hash  : Hash.t
+           ; }
+
+  type error = Invalid_hash of Hash.t
+
+  val pp_error: error Fmt.t
+  val deltas:
+    ?memory:bool
+    -> Entry.t list
+    -> (Hash.t -> Cstruct.t option Lwt.t)
+    -> (Entry.t -> bool)
+    -> int
+    -> int
+    -> ((Entry.t * t) list, error) result Lwt.t
+end
+
 module MakeDelta (H : S.HASH) =
 struct
   module Hash = H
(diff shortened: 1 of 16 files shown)


samoht:
Can you rebase on master now that #272 has been merged? Thanks!

dinosaure:
Cleaned!

src/git/pack.ml
```diff
@@ -53,6 +53,43 @@ struct
     | Tag -> Fmt.pf ppf "Tag"
 end
 
+module type ENTRY =
+sig
```
samoht: minor nitpick: no newline needed here `module type ENTRY = sig`. This is also true in the rest of the patchset :-)

src/git/pack.ml
```diff
@@ -90,9 +90,9 @@ sig
   val topological_sort: t list -> t list
 end
 
-module Entry (H : S.HASH) =
+module MakeEntry (Hash: S.HASH): ENTRY with module Hash = Hash =
```
samoht: I think it prefer `Entry` instead of `MakeEntry`

src/git/pack.ml
```diff
@@ -292,9 +292,9 @@ sig
   val used_out: t -> int
 end
 
-module MakeHunkEncoder (H : S.HASH) =
+module MakeHunkEncoder (Hash : S.HASH): H with module Hash = Hash =
```
samoht: I think I prefer `Hunk_encoder` instead of `MakeHunkEncoder`

src/git/pack.ml
```diff
@@ -641,10 +641,14 @@ sig
     -> ((Entry.t * t) list, error) result Lwt.t
 end
 
-module MakeDelta (H : S.HASH) =
+module MakeDelta
```
samoht: I think I prefer `Delta` instead of `MakeDelta`

src/git/pack.ml
```diff
@@ -975,19 +975,25 @@ sig
   val eval : Cstruct.t -> Cstruct.t -> t -> [ `Flush of t | `Await of t | `End of (t * Hash.t) | `Error of (t * error) ]
 end
 
-module MakePACKEncoder
-    (H : S.HASH with type Digest.buffer = Cstruct.t)
-    (D : S.DEFLATE)
-  : ENCODER with module Hash = H
-             and module Deflate = D =
+module MakePackEncoder
```
samoht: `Pack_encoder` ?

src/git/pack.ml
```diff
@@ -1552,3 +1552,15 @@ struct
     ; hash = Hash.Digest.init ()
     ; state = (Header (header objects)) }
 end
+
+module MakeStreamEncoder
```
samoht: `Stream` ?

src/git/unpack.ml
```diff
@@ -2212,3 +2212,25 @@ struct
...
+    (Hash: S.HASH)
+    (Inflate: S.INFLATE)
+= struct
+  module HunkDecoder = MakeHunkDecoder(Hash)
+  module PackDecoder = MakePackDecoder(Hash)(Inflate)(HunkDecoder)
+
+  include PackDecoder
+end
+
+module MakeRandomAccessPACK
```
samoht: `Random_access` ?

src/git/unpack.ml
```diff
@@ -2131,60 +2155,82 @@ struct
...
     | Error exn -> Lwt.return (Error exn)
     | Ok length ->
       let tmp = Cstruct.create length, Cstruct.create length, length in
 
-      optimized_get' ?chunk ?h_tmp t absolute_offset tmp z_tmp z_win
+      get_from_offset ?chunk ?htmp t absolute_offset tmp ztmp zwin
+end
+
+module MakeStreamDecoder
+    (Hash: S.HASH)
```
samoht: `Stream`?

samoht:
Thanks for the changes (and nice new moustache! :p)

It seems like a nice change, but could show an example on why this is better for the end-user (I guess having flat modules equalities instead of nested one is a nice advantage already).

The only minor comments that I made are about names: I still feel that most of the names could be clean-up: use of snake-case, remove of unnecessary prefix (`Make`) or suffixe (`Encoder`/`Decoder`). Would be nice to clean that a bit before merging that PR.

More generally I think we could still make the API a bit "lighter": for instance, instead of having 6 `get_from_*` and `get_by_*` functions in `unpack.mli`, try to use optional arguments (this could be done in an other PR). An other thing (which also could be done in another PR) is to avoid duplicating signatures in `mli` and `ml` by putting the signatures into an`foo_intf.ml` file and doing `include Foo_intf` in both the `ml` and `mli`.

dinosaure:
So about the best example to show why we need this PR is to use `Inflate` directly instead `PACKDecoder.Inflate` (at this [line](https://github.com/mirage/ocaml-git/pull/273/commits/f5f4d2bc40620306c836753bf3676575631b5d84#diff-2aa69115584784ed1c600cc8387bcddeL43)). Indeed, when we flat inner modules with destructive substitution, we keep the equality between inner types and replace them by the first definition.

So, in this PR, we follow a [pattern](https://bpaste.net/show/b17691d9cbcb) from `Pack`/`Unpack` to `Mem`/`Store` (we have 4 levels of functors for `Unpack`). Finally, in `Store`/`Mem`, we are free to use `PackDecoder.Inflate` or `Inflate` and OCaml stills happy with that.

Previously this PR, we did a hard game between types and resulting modules from functors and it's very unpleasant to have a type error from structurally equal types.

hannesm:
> An other thing (which also could be done in another PR) is to avoid duplicating signatures in mli and ml by putting the signatures into anfoo_intf.ml file and doing include Foo_intf in both the ml and mli.

hmm, I'm usually struggling with this for several reasons: the tooling (i.e. merlin, doc generation?) don't deal nicely with `include` (at least when I try it), it is hard to find definitions (using e.g. `grep`).  these days I elude the problem by not defining any signature in the `.ml` - would this work here as well?

samoht:
@hannesm that's a good point, but I think merlin is working fine now with `include` statements (need to double-check). I am also not super-sure it is a very good idea, but duplicating the module types is also pretty annoying...

src/git/unpack.mli
```diff
@@ -720,17 +720,15 @@ module MakeDecoder
 
 module MakeStreamDecoder
     (Hash: S.HASH)
-    (Inflate: S.INFLATE)
-: sig
```
samoht: you can keep as it was before here, as you already started to cut the line (or put everything on the same line).

Yea I know, someday I should really contribute to `ocamlformat` ...

src/git/pack.ml
```diff
@@ -292,7 +292,7 @@ module type H = sig
   val used_out: t -> int
 end
 
-module MakeHunkEncoder (Hash : S.HASH): H with module Hash = Hash =
+module Hunk_encoder (Hash : S.HASH): H with module Hash = Hash =
```
samoht: do you need the `_encoder` suffix here?

src/git/pack.ml
```diff
@@ -975,24 +975,24 @@ module type P = sig
   val eval : Cstruct.t -> Cstruct.t -> t -> [ `Flush of t | `Await of t | `End of (t * Hash.t) | `Error of (t * error) ]
 end
 
-module MakePackEncoder
+module Pack_encoder
```
samoht: do you need the `_encoder` suffix here?

hannesm:
> @hannesm that's a good point, but I think merlin is working fine now with include statements (need to double-check). I am also not super-sure it is a very good idea, but duplicating the module types is also pretty annoying...

I just tried this: opened in an emacs buffer `ocaml-tls/mirage/example2/unikernel.ml`, `M-x merlin-use<ret>tls.mirage cohttp-lwt cohttp-mirage`, `merlin-locate` on line 33 `Http.make` (the `make` itself is where my cursor is) -> jumped to line 11 of same file, `merlin-locate` on `Cohttp_mirage.Server` (the `Server` part) -> jump to `~/.opam/4.06.0/lib/cohttp-lwt/make.ml` line 3 (`include Cohttp.Request`) -- now I can't move further, merlin says `"Not in environment 'Cohttp.Request'"`.  It did follow the module alias `module Server = Make.Server` in cohttp_mirage.ml (and opened `make.ml`).

avsm:
I think that's a bug in Merlin.  Shouldn't we fix the tools rather than constraining the module design?  We can also look into addressing it in Odoc to make sure it handles odoc ok.

samoht:
We should definitely fix the tool, but that change is outside the scope of that PR, as we already duplicate module signatures. The goal os this PR is just to flatten the submodules of these already existing signatures.

PR #274, opened 2018-01-18
Indentation & style

dinosaure:
Need #272 and #273 (not need to merge)

PR #275, opened 2018-01-18
Some final stuffs (most about style)

dinosaure:
Need #272, #273 and #274 (not need to merge). More about @samoht's style.

### DIFF
--- a/src/git/pack_engine.ml
+++ b/src/git/pack_engine.ml
@@ -776,8 +776,8 @@ module Make
          | Ok ()          -> sys_err
          | Error sys_err' ->
            Log.err (fun l ->
-               l "Impossible to close the fd of the PACK file (to analyze it): \
-                  %a." FS.pp_error sys_err');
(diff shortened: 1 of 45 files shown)


src/git/pack_engine.ml
```diff
@@ -776,8 +776,8 @@ module Make
          | Ok ()          -> sys_err
          | Error sys_err' ->
            Log.err (fun l ->
-               l "Impossible to close the fd of the PACK file (to analyze it): \
-                  %a." FS.pp_error sys_err');
+               l "Impossible to close the fd of the PACK file (to analyze it), \
+                  ignoring: %a." FS.pp_error sys_err');
```
samoht: Do we really want to display that here, or should that message simply be in `pp_error` and we let the user print the error if needed? I think it's better to not print anything as long as the error is not discarded and let the use decide later to print/ignore the error.

dinosaure:
Lazy to rebase it.

PR #277, opened 2018-02-20
Low-level part of server implementation

dinosaure:
This PR is a big clean-up (with tests) of the low-level server part of `ocaml-git`. It provides:

* A better interface on the Smart protocol: we have a `Common` part shared by `Decoder` and `Encoder`, it defines common OCaml values on Smart protocol
* A fuzzer (available with `jbuilder build @runfuzz`), not fully implemented and not fully launched (very long time) to test if `Decoder.of_string (Encoder.to_string) v = v` (isomorphism)
* Tests about some semantics about what we expect as an input and as an output in depending on capabilities.

For the last point, it's about some part of Smart protocol which the specification change from shared capabilities between the client and the server - for example, negotiation phase changes if we use `multi-ack-detailed` or `multi-ack` capability.

Do not merge ... not finish.


dinosaure:
Travis is happy (but revdeps catched a problem) and AppVeyor has a problem with `camlzip`.

avsm:
cc @yomimono - more fuzzing !

dinosaure:
Yeah, the fuzz part is available on the `fuzz/` directory. The `rabin` fuzz is to see if we produce a good _delta-ification_ and reconstruct object by it without divergence.

The smart fuzz is about the protocol Smart and if `Decode.of_string (Encode.of_string v) = v` at any times.

However, as I said, because the fuzzer is very slow I did not launch a long time and just catch some little bugs - so I did not use `ocaml-bun`.

dinosaure:
So, this PR is most finished (the HTTP overlay need to be check) but the low-level part of the Smart protocol (I mean the implementation of the protocol itself) is done and I can focus on the high-level part, I mean the logic behind the server (negotiation engine, how to store a PACK, etc.).

However, I did not run an AFL test with afl-fuzz. Firstly because my computer will die and secondly because I don't have a time to wait results from the fuzzer. But all part of the Smart protocol has a fuzzer (provided by `jbuilder build @runfuzz`). I will launch the fuzzer on one of my server a look later results.

yomimono:
If you're interested in having Travis run the fuzzer for you, they're (for some reason) happy to give you large amounts of free CPU time as long as your program continues to put stuff on stdout - see https://github.com/yomimono/ocaml-test-stdlib/blob/primary/.travis.yml for an example of how to set this up (it uses `bun` but you don't have to, although then you'll either only get one CPU or need to figure out how to launch multiple instances).

yomimono:
(actually, better to look at https://github.com/yomimono/ocaml-test-stdlib/blob/crowbar_abstract_gens/.travis.yml , which doesn't have so much cruft in it)

samoht:
What's the status of that branch?

dinosaure:
Ready to merge I believe.

samoht:
Could you rebase to remove the merge commits?

dinosaure:
Cleaned :+1:

dinosaure:
This PR was not merged for a long time because CI fails all the time about `tls`/`ssl` support. However it integrates lot of tests about the smart protocol. The only commit which should be revert (but situation about `tls`/`ssl` is not clear) is 54d7862.

So if CI is _happy_, I will merge it.

samoht:
You don't need to revert the TLS/SSL patch - my fix on windows is to not link with tls and git-unix (will be fixed in 1.0.2). Will maybe need to force the installation of `ssl` on the windows CI though..

dinosaure:
Hmmhmm, I can not reproduce error on my Windows about `appveyor`, it's really weird ...

dinosaure:
Ok error on windows is available [here](https://ci.appveyor.com/project/samoht/ocaml-git-7vmk0/build/1.0.691/job/ip1j4or4s9k4aynm#L15199). I clearly don't understand what is the deal between `/` and `\` on windows with (or without?) `cygwin` (the execution context of tests is not clear for me on `appveyor`)

Currently, this path is printed with `fpath` which integrates as global a variable to know if it's needed to use `\` instead `/`. I can trick on the code to use `/` (and this trick should fix CI) but if someone has a better way (@dra27 or @nobj) ...

dinosaure:
Ok `git-mirage` is happy on `appveyor` with my hack. MERGE IT!

PR #280, opened 2018-03-01
Fix angstrom

dinosaure:
Update `ocaml-git` to the last API of `angstrom`.

PR #279, opened 2018-02-25
Fix commit

dinosaure:
Fix #276 and handle extra-fields.

PR #281, opened 2018-03-01
Add iter

dinosaure:
Add iter function (special case of `fold` in `traverse_bfs.ml`). Same as #279, wait to merge #280 and #279, clean up this PR and ready to merge.


dinosaure:
Bad history, see #283.

PR #287, opened 2018-03-27
Encore integration

dinosaure:
This is a big PR to use [`encore`](https://github.com/dinosaure/encore.git) instead _decoder_ and _ encoder_ wrote by hands. This PR fixed a problem about `tree` object (see the commit about that). We delete `faraday` (and use the internal encoder of `encore`) and all seems to work.

Not ready to merge, we need to wait release of `encore` (which needs an improvement of API).

A note about reference, now, we expect a `\n` at the end of `head_contents` (before it's optional).


dinosaure:
+1,117 −2,326 - good ratio 👍

samoht:
It's a bit hard to review but as all the tests still pass I guess it is ok :-)

PR #289, opened 2018-04-03
Replace Radix by Stdlib Map module

dinosaure:
About performance, Radix is slower on add compare to Map and
faster on lookup (~13 %). It's clearly not really interesting to keep
(and maintain it) instead to use Map.


dinosaure:
Cleaned!

test/common/test_common.ml
```diff
@@ -123,13 +123,7 @@ module Make (S: Git.S) = struct
...
       (Fmt.Dump.pair S.Reference.pp S.Hash.pp)
 
-  module Radix =
-    Git.Radix.Make(struct
-      type t = S.Hash.t
-
-      let get = S.Hash.get
-      let length _ = S.Hash.Digest.length
-    end)
+  module Radix = Map.Make(S.Hash)
```
samoht: Still some Radix here :-)

PR #290, opened 2018-04-05
Abstract file-system

dinosaure:
This PR is a step to abstract step by step file-system and limit POSIX syscall on the git core library. Currently, the status is to use `to_file` and `of_file` which use a subset of POSIX syscall (and can be easily translated to another back-end like [wodan](https://github.com/g2p/wodan)) to load or store a Git object (loose or reference) from a file-system.

However, this abstraction is provided by the `Helper` module which stills on the Git core library.

Then, pack-engine stills to use POSIX syscall in only one place (`Pack_engine`) to do a first pass but an update of the pack-engine is out of the scope of this PR - and compatibility between `wodan` and the pack-engine stills an open question.

### DIFF
--- a/src/git/loose.ml
+++ b/src/git/loose.ml
@@ -199,11 +199,11 @@ module Make
           | Error _ -> acc)
         [] firsts
 
-  type ('result, 'state) decoder =
-    (module Helper.DECODER with type t = 'result
-                            and type state = 'state
-                            and type error = [ Error.Decoder.t
-                                             | `Inflate of Inflate.error ] )
+  type ('result, 'decoder) decoder =
+    (module Helper.DECODER
+            with type t = 'result
+             and type decoder = 'decoder
+             and type error = [ Error.Decoder.t | `Inflate of Inflate.error ] )
 
   let gen (type state) (type result) ~fs ~root (state:state) ~raw (decoder : (result, state) decoder) hash =
     let module D = (val decoder) in
@@ -221,15 +221,6 @@ module Make
 
   let read ~fs ~root ~window ~ztmp ~dtmp ~raw hash =
     let state = D.default (window, ztmp, dtmp) in
-    let module D = struct
-        type state = D.decoder
-        type error = D.error
-        type t = D.t
-
-        let eval = D.eval
-        let refill = D.refill
-        let finish = D.finish
-      end in
     gen ~fs ~root state ~raw (module D) hash
 
   module HeaderAndBody = struct
@@ -285,15 +276,6 @@ module Make
 
   let inflate ~fs ~root ~window ~ztmp ~dtmp ~raw hash =
     let state = I.default (window, ztmp, dtmp) in
-    let module I = struct
-        type state = I.decoder
-        type error = I.error
-        type t = I.t
-
-        let eval = I.eval
-        let refill = I.refill
-        let finish = I.finish
-      end in
     gen ~fs ~root state ~raw (module I) hash
 
   let inflate_wa ~fs ~root ~window ~ztmp ~dtmp ~raw ~result hash =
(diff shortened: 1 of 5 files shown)


src/git/loose.ml
```diff
@@ -199,47 +199,38 @@ module Make
...
+    let module D = struct
+        type state = D.decoder
+        type error = D.error
+        type t = D.t
+
+        let eval = D.eval
+        let refill = D.refill
+        let finish = D.finish
+      end in
+    gen ~fs ~root state ~raw (module D) hash
```
samoht: why do you need to build an intermediate `D` module here? I think you could just pass `(module D)` and it should work.

src/git/loose.ml
```diff
@@ -199,47 +199,38 @@ module Make
...
+       Lwt.return Error.(v @@ Error.Decoder.with_path file err)
+    | Error (`Decoder (`Inflate _ as err)) ->
+       Lwt.return Error.(v @@ Inf.with_path file err)
+    | Error #fs_error as err -> Lwt.return err
 
   let read ~fs ~root ~window ~ztmp ~dtmp ~raw hash =
-    gen ~fs ~root ~window ~ztmp ~dtmp ~raw (module D) hash
+    let state = D.default (window, ztmp, dtmp) in
+    let module D = struct
+        type state = D.decoder
```
samoht: can we simplify this? e.g. by renaming `type state` to `type decoder` in some places (or the other way around)

samoht:
That looks like a very big improvement, but there is still a bit of duplication and verbosity. Could we try to consistently use `state` or `decoder` to avoid having to create these dummy `D` modules everywhere?

src/git/loose.ml
```diff
@@ -199,47 +199,29 @@ module Make
...
-            Lwt.return Error.(v @@ Inf.with_path path err)
-    in
-    loop decoder >|= fun ret ->
-    Log.debug (fun l -> l "Finished to read the object %s / %s." first rest);
-    ret
+    let file        = Fpath.(root / "objects" / first / rest) in
+    Log.debug (fun l -> l "Reading the loose object %a." Fpath.pp file);
+    Decoder.of_file fs file raw state >>= function
+    | Ok v -> Lwt.return (Ok v)
+    | Error (`Decoder (#Error.Decoder.t as err)) ->
```
samoht: do you really need to specify the error type here? Would something like work?

```
| Error (`Decoder err) -> Error (v @@ Error.Decoder.with_path file err)
```

src/git/packed_refs.ml
```diff
@@ -155,25 +155,17 @@ module Make (H: S.HASH) (FS: S.FS) = struct
...
-          match D.refill (Cstruct.sub raw 0 n) decoder with
-          | Ok decoder              -> loop decoder
-          | Error (#Error.Decoder.t as err) ->
-            Lwt.return Error.(v @@ Error.Decoder.with_path path err)
-    in
-    loop decoder
+    let state = D.default dtmp in
+    let file  = Fpath.(root / "packed-refs") in
+    Decoder.of_file fs file raw state >>= function
+    | Ok v -> Lwt.return (Ok v)
```
samoht: same story here

src/git/reference.ml
```diff
@@ -303,34 +305,13 @@ module IO (Hash: S.HASH) (FS: S.FS) = struct
...
-    | Ok head_contents ->
-      let reference' = normalize path in
-      Log.debug (fun l ->
-          l "Normalize reference %a = %a." pp reference pp reference');
-      assert (equal reference reference');
-      Ok (normalize path, head_contents)
+    let state = D.default dtmp in
+    let file  = Fpath.(root // (to_path reference)) in
+    Decoder.of_file fs file raw state >>= function
+    | Ok v -> Lwt.return (Ok (normalize file, v))
```
samoht: again

src/git/loose.ml
dinosaure: Only on this case (about `packed-refs` and `reference`, we don't have this problem), `Decoder` can have 2 kinds of error:
- `Inflate`: about `inflate` implementation
- `Error.Decoder.t`: usual decoder error

OCaml seems to be not so clever about pattern-matching and polymorphic variant. So, if I don't use `#Error.Decoder.t`, it includes `Inflate` error. However, `Error.Decoder.with_path` does not expect this kind of error. So, yeah :( I need to refine pattern-matching.

About `packed-refs` and `reference`, because we don't have `Inflate` error, I can change it, of course (more readable).

PR #291, opened 2018-04-13
Fix test-and-set for packed references

samoht:
As ocaml-git 1.11 never unpack the packed references (instead it just writes
loose references when it wants to update them), test_and_set_reference needs
to be adapted.

Note: remove_reference and test_and_set ~set:None are completely buggy when
using packed references. This will be fixed by ocaml-git 2.0 but there is
no easy fix in ocaml-git 1.11... So simply add a note in API docs.

### DIFF
--- a/src/git/fs.ml
+++ b/src/git/fs.ml
@@ -22,8 +22,6 @@ open Astring
 module PackedLog = (val Misc.src_log "fs-packed" : Logs.LOG)
 module LooseLog = (val Misc.src_log "fs-loose" : Logs.LOG)
 
-module ReferenceSet = Misc.Set(Reference)
-
 let failf fmt = Fmt.kstrf failwith ("Git.FS." ^^ fmt)
 
 let err_not_found n k = failf "%s: %s not found" n k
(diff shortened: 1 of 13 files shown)


src/git/fs.ml
```diff
@@ -635,7 +642,13 @@ module Make (IO: IO) (D: Hash.DIGEST) (I: Inflate.S) = struct
       | Some v -> Some (Cstruct.of_string (Hash.to_hex v))
     in
     let lock = lock_file t r in
-    IO.test_and_set_file file ~temp_dir ~lock ~test:(raw test) ~set:(raw set)
+    let is_read_packed_ref = Reference.Set.mem r t.read_packed_refs in
+    let test = if is_read_packed_ref then None else raw test in
```
talex5: Does this mean that when updating a packed ref, you can specify *any* value for `test` and it will update? I guess if this code is going away soon it probably doesn't matter, but might be worth a warning comment in the API.

samoht: This should be fixed now.

PR #292, opened 2018-04-30
New pack engine

### LINKED ISSUE #286: FS.create is too slow to use as a command line client
Would it be possible to have a fast version of ```FS.create``` for such uses ?
I guess one way would be to have a ```FS.create_client``` or ```FS.create_server``` instead maybe ? Or something like ```FS.longterm_create``` and ```FS.create``` ? I don't know what would be the best interface here but that's just my two cents.

dinosaure:
So this PR is a start to fix #286 and:

* Have a pack-engine which can be used in a context server or as a client
* Let the user to handle allocation

Firstly, we change a little bit the unpack module to be more general than before. The biggest change is to use an ADT to use a string or noticed `Cstruct.t` by the user instead to allocate some `Cstruct.t` (of 127 bytes) when the user did not noticed a already allocated `Cstruct.t`.

This `Cstruct.t`/`string` is used to store insert hunks while we delta-ify an object - and keep them at each level of delta-ification to apply them then at the end of the recursion. In _sirodepac_, I already noticed than `string` is not the best buffer to store them (specially when we wants to extract all objects of a PACK file, the OCaml GC does not support the pressure) but the solution with `Cstruct.t` is not good too for allocate a so tiny buffer (127 bytes is the maximum). Both seem to be the solution ...

Then, from @samoht's advise and benchmark, `Pack_info` switch to use `Hashtbl` instead a `Map`. Then, a flash of genius under my shower gave me a solution to resolve delta-ification from a already existed PACK file - the solution is to use the IDX file to resolve `OBJ_REF_DELTA` and compute what is needed to reconstruct objects.

Indeed, when an object is delta-ify, the source can be noticed by `OBJ_OFS_DELTA`, in this case, we just follow offset but it could be delta-ify by an object referenced by `OBJ_REF_DELTA` which is the hash of the source. The problem is, if the source object is delta-ified too, we did not compute hash from it yet (a second pass is necessary). However, with the IDX file, we can retrieve this object and if we did not find the source, that means the PACK file is thin.

This is the big change about `Pack_info`.

Finally, we can talk about the `Pack_engine` and some changes:

- First, we decide to cut in some states the `Pack_engine`:
  * `exists`: when the PACK file already exists on the git repository
  * `loaded`: when we load the PACK file (we promote `exists` to `loaded` when the user wants to read an object from this PACK file - but `lookup`, `list` and `mem` operations don't do this promotion)
  * `normalized`: when the PACK file comes from a stream (when you `pull`/`fetch`/`clone`)
  * `resolved`: when we resolved all delta-ified objects in the PACK file
  * `total`: when we ensure than the PACK file is _thin_ or not

As you can see, we can have two entry point for a PACK file:
- if the PACK file already exists on the git repository
- if the PACK file comes from a stream

As `git`, if the PACK file comes from a stream, we will apply 3 passes. The first one can be done while we store it on a temporary directory. Then, the second pass will try to resolve all delta-ified objects from this PACK file (always stuck on the temporary directory). With the second pass, we can determine if the PACK file is _thin_ or not. In the second case, we just do an atomic move from temporary directory to git repository and generate the IDX file. In the first case, we generate a new PACK file to ensure than any PACK file in the git repository are non-_thin_ (like what `git` does).

Obviously, this computation is little bit slow, however, then, we add a new PACK file from a stream directly as a `total` PACK file. That means, we calculated needed `Cstruct.t` to extract any objects from this PACK file and control allocation at this stage.

Finally, for an already existed PACK file, we did not do this kind of computation - but we don't control allocation - and when we have enough to promote it to a new state (`resolved`), we do it. Finally, in common way, for the next call, the PACK file will be promoted as a `total` PACK file (because any PACK file in the git repository is _thin_).

So, in this way, we did not do a pass automatically (and scan all PACK file) on already existed PACK file. However, the client will be able to force this scan if he wants - but it's not the default behavior now and should be fix #286.

- Secondly, we let the user to control allocations

This is the biggest mandatory for me, it's to let the user to keep the control about the allocation. I created a new record:

```ocaml
type ('location, 'mmu) r =
   { mmu              : 'mmu
   ; with_cstruct     : 'mmu -> int -> (('location * Cstruct.t) -> unit Lwt.t) -> unit Lwt.t
   ; with_cstruct_opt : 'mmu -> int option -> (('location * Cstruct.t) option -> unit Lwt.t) -> unit Lwt.t
   ; free             : 'mmu -> 'location -> unit Lwt.t }
```

Which let the user to define a politic about allocation for `Cstruct.t` in some context. Obviously, we have an easy context where the user can just allocate one `Cstruct.t` which needs to grow specially when we apply the second pass. Then, the second context is, as @samoht asked, a concurrent context when we calculated what is needed to extract any objects from the PACK file (so `resolved` and `total` state). In this context, we will ask a fixed size and buffer will be use on a concurrent context (and it's the purpose of the `free` function which could be a Lwt waiter).

Then, I did some check (by hand) to ensure ownership on `Cstruct.t` but, as before, we delimited allocation of `Cstruct.t` in only one case: `read_and_exclude`.

- Third, we fixed `read_and_exclude`

Indeed, in a special context, a PACK file A can have an delta-ified object O which needs an object M outside from the PACK file A. The source object could be appear on a PACK file B but nothing in the documentation tell me than this object __can not__ be delta-ified by O. So we needs to avoid a circular dependency and because we can find multiple times the same object (like O) in multiple PACK files, we need to check to take object O from another PACK file than A.

This is the purpose of `read_and_exclude`. The code looks like a deep and non-optimized code which allocate a lot, and it's true but it's the worst comes to the worst - and from my knowledge, it should not appear ...

- Fourth, we replace `Map` by `Hashtbl` like `Pack_info`

This PR is not finished, we need to merge this store on `Store` and `Mem` now but the interface does not change a lot and should delete some duplicate codes (specially in `Store`). But I make this PR to follow development and get some advises before merging. So feel free to ask something if you want!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ _build/
 *~
 *.install
 .merlin
+_opam
+.envrc
--- a/src/git/index_pack.ml
+++ b/src/git/index_pack.ml
@@ -38,6 +38,7 @@ module type LAZY = sig
   val mem: t -> Hash.t -> bool
   val iter: t -> (Hash.t -> (Crc32.t * int64) -> unit) -> unit
   val fold: t -> (Hash.t -> (Crc32.t * int64) -> 'a -> 'a) -> 'a -> 'a
+  val cardinal: t -> int
 end
 
 module Lazy (H: S.HASH): LAZY with module Hash = H = struct
@@ -129,6 +130,8 @@ module Lazy (H: S.HASH): LAZY with module Hash = H = struct
        ; values_offset
        ; v64_offset
        ; cache = Cache.create ~random:true cache }
+
+  let cardinal { number_of_hashes; _ } = number_of_hashes
 
   exception Break
 
--- a/src/git/index_pack.mli
+++ b/src/git/index_pack.mli
@@ -68,6 +68,8 @@ module type LAZY = sig
 
   val fold: t -> (Hash.t -> (Crc32.t * int64) -> 'a -> 'a) -> 'a -> 'a
   (** Fold in the IDX file. *)
+
+  val cardinal: t -> int
 end
 
 module Lazy (H: S.HASH): LAZY with module Hash = H
--- a/src/git/mem.ml
+++ b/src/git/mem.ml
@@ -253,7 +253,7 @@ module Make (H: S.HASH) (I: S.INFLATE) (D: S.DEFLATE) = struct
         let contents _ = assert false
       end)
 
-    module Graph = GC.Graph
+    module Map = GC.Graph
 
     let make = GC.make_stream
 
--- a/src/git/minimal.ml
+++ b/src/git/minimal.ml
@@ -132,7 +132,7 @@ module type S = sig
     type stream = unit -> Cstruct.t option Lwt.t
     (** The stream contains the PACK flow. *)
 
-    module Graph: Map.S with type key = Hash.t
+    module Map: Map.S with type key = Hash.t
 
     val from: t -> stream -> (Hash.t * int, error) result Lwt.t
     (** [from git stream] populates the Git repository [git] from the
@@ -144,7 +144,7 @@ module type S = sig
       -> ?window:[ `Object of int | `Memory of int ]
       -> ?depth:int
       -> Value.t list
-      -> (stream * (Crc32.t * int64) Graph.t Lwt_mvar.t, error) result Lwt.t
+      -> (stream * (Crc32.t * int64) Map.t Lwt_mvar.t, error) result Lwt.t
     (** [make ?window ?depth values] makes a PACK stream from a list
         of {!Value.t}.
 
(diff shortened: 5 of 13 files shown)


dinosaure:
Ah yeah, comments in `Pack_engine` is in french, sorry ...

dinosaure:
And of course, CI will fail because the code does not compile yet.

dinosaure:
Ready to merge.

samoht:
I have pushed some translation of the comments from english to french and added a few tweaks to the API (removal of `with_cstruct_opt`, renaming of a few functions to make clearer what they do). 

I think there is still room to improve the API a bit: the name of the state are not really related to what are, and they are a few more assumptions about what these states are which are not cleanly stated yet. But this could be done in an other PR.

PR #293, opened 2018-05-04
Make Tree.entry a private type

samoht:
This allows to check that steps are well-formed only once; add also convenient
fonctions to efficiently add/remove entries in a tree.


samoht:
The errors are due to `qtest` archive being unavailable.

dinosaure:
I'm ok to merge.

PR #296, opened 2018-05-07
Replace Rabin by duff package

dinosaure:


samoht:
Would have been simpler to have called that new package `Xdiff` :-) 

Anyway, LGTM

PR #297, opened 2018-05-15
Fix compilation and replace Rabin by Duff in Pack_engine module

dinosaure:
Missing a replacement.


dinosaure:
Fail by ocaml/opam-repository#11991

PR #300, opened 2018-06-20
Second pass

dinosaure:
A PR to prepare to the new PR which will implement an ascendant walker of PACK file (and, btw, improve performances).


samoht:
LGTM

I'll try to fix the CI.

PR #301, opened 2018-06-21
Try to fix appveyor

samoht:


samoht:
The tmp issues should be fixed be https://github.com/mirage/ocaml-git/pull/305

PR #303, opened 2018-07-09
Move to the last version of decompress (released)

dinosaure:


dinosaure:
We break the API (already) and `ppx_tools_versioned` seems to have a problem. In other case, this PR can be merged - where I did not change the semantic of `decompress`.

samoht:
The ppx_tools_versioned error should be fixed by https://github.com/ocaml/opam-repository/pull/12348

PR #305, opened 2018-08-07
Simplify the use of temporary directory (for atomic renames)

samoht:
Always use `.git/tmp` to avoid isues with cross-device links.


samoht:
CI on windows is broken because of:

```
checkseum is not available because your system doesn't comply with ocaml-version >= "4.03.0" & os != "win32".
```

which was due to https://github.com/fdopen/opam-repository-mingw/commit/857a012189c005ac2c0aef4c96f64abf990a12ab

Merging this to fix the CI on Linux at least.

PR #306, opened 2018-08-07
Add a few mlis and remove dead-code

samoht:
The annoying mli to add is `error.mli`: I have a half-patch for this, but it is not complete yet and I am not sure I am brave enough to push it through completion.


samoht:
windows CI is dead to to checkseum not workin on windows ... will be fixed later.

PR #308, opened 2018-08-31
Expose feed_b and feed_s in Git.Hash.S to avoid copying too much cstr…

samoht:
…uct contents


dinosaure:
We still have a problem on `appveyor` but it seems good.

samoht:
Yes I hope that https://github.com/ocaml/opam-repository/pull/12552 will finally fix the issue with windows/appveyor.

PR #314, opened 2018-09-13
Common sync

dinosaure:
This purpose of this PR is to be able to make a switch _à la conduit_ from an URI and choose automatically the TCP back-end (`git` package) or HTTP back-end (`git-http`). Now, these implementations share the same interface `Git.Sync.S` which provide high level API and low level API.

Endpoint (URI or URI with headers) is abstracted on this interface.

PR is huge and this biggest addition/deletion is a move of the TCP implementation from `Git.Sync` to `Git.Tcp`. Now, `Git.Sync` exports only common interface (`S` and `ENDPOINT`) and common implementation about some operations on the store when we `fetch` or `push` (independently from network layer - this is only a module about _local_ operations).

### DIFF
--- a/test/common/dune
+++ b/test/common/dune
@@ -1,4 +1,4 @@
 (library
   (name      test_git)
   (wrapped   false)
-  (libraries git mtime mtime.clock.os alcotest nocrypto lwt.unix logs.fmt))
+  (libraries git git-http mtime mtime.clock.os alcotest nocrypto lwt.unix logs.fmt))
--- a/test/common/dune
+++ b/test/common/dune
@@ -1,4 +1,4 @@
 (library
   (name      test_git)
   (wrapped   false)
-  (libraries git git-http mtime mtime.clock.os alcotest nocrypto lwt.unix logs.fmt))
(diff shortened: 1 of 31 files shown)


test/common/dune
```diff
@@ -1,4 +1,4 @@
 (library
   (name      test_git)
   (wrapped   false)
-  (libraries git mtime mtime.clock.os alcotest nocrypto lwt.unix logs.fmt))
+  (libraries git git-http mtime mtime.clock.os alcotest nocrypto lwt.unix logs.fmt))
```
samoht: some tests have to work without depending on `git-http` (e.g the git in-memory one for instance).

samoht:
That looks good. I don't particularly like the `MakeCohttp` name, but I have no better suggestion... Do you really need to expose that functor instead of just have `Make`?

The "common" tests need to compile without depending on git-http.

Also, what is your strategy for allowing users to connect to either `git://` or `http://` or `git+ssh://` with the same sync function?

dinosaure:
Currently, I did this PoC: https://gist.github.com/dinosaure/b7a8ebf42811836e6a9a3ae299436673

PR #315, opened 2018-09-20
Expose Sync.HTTP in git-unix and git-mirage

samoht:


dinosaure:
I'm currently working on it to put the dispatcher but it seems than we don't have a strong conflict - only difference is the way to provide implementations on `git-unix` and `git-mirage` but it seems fine for me.

dinosaure:
After discussions, this solution seems the best about `irmin` and resolution of uri's scheme (`git://`, `http://` or `https://`). Just need to add `git+ssh://` which needs to update ADT but we don't need to do something more extensible than it.

PR #318, opened 2018-09-25
unix: be less verbose on expected Unix errors

samoht:
Also precise the function name when the error is a real one.


samoht:
CI issue is fixed by https://github.com/mirage/ocaml-git/pull/319, so merging.

PR #320, opened 2018-09-25
Replace eq with equal

zshipko:
ocaml-git isn't building due to `eq` being renamed to `equal`.  This can also be observed in the CI logs for irmin: https://github.com/mirage/irmin/pull/552.

Feel free to close this if it's already being dealt with!


dinosaure:
Thanks 👍 waiting CI and merge it!

PR #325, opened 2018-10-06
Remove Git_mirage.Store

samoht:
That's not the way we want to support Git backend with Mirage: there's
too many POSIX assumptions baked in that model...


samoht:
No idea what the unix errors mean... /cc @dinosaure

PR #328, opened 2018-10-12
Index windows

dinosaure:
Fix finally INDEX file on Windows ... (this PR is on top of #327)


samoht:
Maybe call it `Git.Path` instead?

PR #336, opened 2018-12-30
Allocation

dinosaure:
Big pass to be able to use an already allocated buffer to encode Git object. This PR was thinked at the beginning of `ocaml-git.2.0.0` but it was never apply ... This is should optimize `ocaml-git` when I deleted most of systemic allocation - however, we still have some allocations in some place but I figure out currently on this issue.


dinosaure:
Ah and some noise about documentation ...

dinosaure:
`ke` is not available yet on Windows but it's fine. I merge it!

PR #339, opened 2019-01-23
Remove digestif.{c,ocaml}, checkseum.{c,ocaml} dependencies on git-unix and git-mirage libraries

dinosaure:
This PR is like a PR out of box but from an other project which want to link with `irmin-unix` and `irmin-mirage`, it wants to link with `git-unix` and `git-mirage` by transitivity. However, both compiles with `digestif.c` & `digestif.ocaml` (`git-unix`) and `checkseum.c` & `checkseum.ocaml` (`git-mirage`). By this way, we link both implementations.

At the end, we did not retrieve any compiler errors (which is an issue?) but get a _segfault_. This PR should fix this problem.

### DIFF
--- a/src/git-mirage/dune
+++ b/src/git-mirage/dune
@@ -1,6 +1,4 @@
 (library
  (name        git_mirage)
  (public_name git-mirage)
- (libraries   digestif.ocaml checkseum.ocaml
-              git mirage-flow-lwt mirage-channel-lwt
-              mirage-conduit cohttp-mirage git-http decompress))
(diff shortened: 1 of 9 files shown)


hannesm:
thanks for fixing this @dinosaure.  I'm curious whether there is a way to express "never link `.c` and `.ocaml` at the same time" (to avoid potential issues in the future)!?! (or is that already part of dune library variants - will the `.c` and `.ocaml` be replaced by library variants?)

dinosaure:
Yes, indeed it's a part of the dune variant :+1: .

PR #340, opened 2019-01-23
Created simple git log like sample

clecat:


dinosaure:
Due to the new interface of Git, this PR is not compatible with the new version. I will close it but feel free to go back and re-implement it with the new API and then, re-open the PR.

PR #341, opened 2019-01-31
Support for authentication via Uri

linse:
Hello friends! 
We tested the new KV interface with github as a storage backend for caldav. To make it work, we had to add HTTP Basic Auth. With this change, we can authenticate to github and read and write (pull and push). See it working at https://github.com/roburio/testcalendar 
Also, we reverted your change regarding `checkseum.ocaml` and `digestif.ocaml` to make it compile. We tried to add the above libraries as sublibraries in a mirage unikernel's config.ml file, but that did not seem to work yet. 
Enjoy! 🌴


dinosaure:
Hmmhmm about your revert, I will check why your `caldav` does not compile. It seems that #339 hides a much bigger bug than what I expected. I will resolve it next week.

dinosaure:
Close by #343, Thanks!

PR #356, opened 2019-04-17
Use lastest version of lru.0.3.0

dinosaure:


dinosaure:
CI should be not be green with the latest version of `encore`, so merge it to ensure than #355 compiles.

PR #360, opened 2019-05-27
Use mmap package instead Unix.mmap function on unix backend

dinosaure:


dinosaure:
Last commit is unrelated to `mmap` but I take the opportunity to update dependencies of `ocaml-git` and use variants (and, btw, fix compilation of `git-mirage`).

PR #362, opened 2019-06-27
initial ssh (client) support using awa-ssh

hannesm:
ssh.ml is mainly a copy of tcp.ml, some adjustment were needed since in the tcp protocol the client first sends a request, while in ssh a command (git-receive-pack or git-upload-pack) is executed (with the path to the repository as argument) which starts by emitting data.

this means that git-receive/upload-pack needs to be passed while the connection is established (as done with `?cmd` here), i.e. before there's a FLOW (or NET as it is called here).

### DIFF
--- a/git.opam
+++ b/git.opam
@@ -36,8 +36,8 @@ depends: [
   "checkseum"  {>= "0.0.9"}
   "stdlib-shims"
   "ke"
-  "encore"     {>= "0.5"}
-  "duff"
+  "encore"     {>= "0.5" & < "0.6"}
+  "duff"       {< "0.3"}
   "hex"
   "ocplib-endian"
   "rresult"
--- a/src/git-mirage/git_mirage.ml
+++ b/src/git-mirage/git_mirage.ml
@@ -29,6 +29,8 @@ module Sync (G : Git.S) = struct
 
   module Tcp = Git.Tcp.Make (Net) (Endpoint) (G)
 
+  module Ssh = Git.Ssh.Make (Net) (Endpoint) (G)
+
   module Client = struct
     (* XXX(samoht): too much copy/paste from git-unix ... *)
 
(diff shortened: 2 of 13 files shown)


samoht:
That looks great!

Do you think we can avoid duplicating some much code in tcp.ml/ssh.ml. The protocols are mostly similar (apart the message prefixes if I remember correctly) so I hope there is a way to share more code :-)

hannesm:
@samoht I'd hope so, yes. I don't expect this PR to be merged as is, but would appreciate if someone could think about the code and come up with the right abstractions. please take also the dual PR for conduit/mirage-conduit https://github.com/mirage/ocaml-conduit/pull/296 into consideration.

hannesm:
How to use this code in an application?

This PR adds `git+ssh` protocol support to ocaml-git. In order to get a git+ssh client, you'll need as well:
- conduit (basically to handle these `ssh://` urls) with [this PR](https://github.com/mirage/ocaml-conduit/pull/296)
- awa (the actual ssh implementation) from [here](https://github.com/haesbaert/awa-ssh)

For your convenience, the opam repository https://github.com/roburio/git-ssh-dns-mirage3-repo acts as an overlay for git+ssh support (`opam repo add git-ssh https://github.com/roburio/git-ssh-dns-mirage3-repo` should be all you need in your switch).

As application, [the primary DNS unikernel](https://github.com/roburio/unikernels/tree/future/primary-git) is an example, TL;DR: in config.ml it has two new command-line parameters "seed" and "authenticator", in unikernel.ml it stitches this together into a "HTTP header" (that's what irmin and git and conduit API allowed) [see here](https://github.com/roburio/unikernels/blob/future/primary-git/unikernel.ml#L12-L25) and passes this to `Store.remote`. The seed is a random string (use `awa_gen_key` to generate your own) used to produce the private key, the authenticator should be the SHA256 fingerprint of the remote host. As `remote` you can then use `ssh://my-server/my-repo.git` (or `ssh://git@github.com/mirage/ocaml-git.git`).

For a non-MirageOS application, it is likely that the conduit patch has to be applied to conduit-lwt-unix as well.

dinosaure:
Close by #395 which integrate properly SSH.

PR #365, opened 2019-07-08
Check order on tree with of_list/to_list and add operation

dinosaure:
/cc @samoht


dinosaure:
resolution time out about `git-mirage`, In other side, I did not retrieve the bug about `Git.Tree.add` which produces a different order about entries.

samoht:
🤔

PR #371, opened 2019-10-02
Use conduit-mirage instead mirage-conduit

dinosaure:


dinosaure:
Let's merge, an error appear about nocrypto on Windows but I should know how to fix it.

PR #373, opened 2019-10-09
Unlock git to use checkseum.0.0.9 and be able to use MirageOS 3

dinosaure:


dinosaure:
Should be reflected to opam-repository.

PR #376, opened 2019-11-01
adapt to mirageos 3.7.0 changes

hannesm:
CI won't succeed since some dependencies are still not released to opam-repository (but PRed) -- conduit / cohttp


hannesm:
since encore 0.4 is out there now, I restarted the travis job here -- and it succeeded! so imho this should be good to be merged and released to opam-repository! /cc @dinosaure

dinosaure:
Thanks!

PR #378, opened 2019-11-14
Fix bug about padding needed for an index file (version 2)

dinosaure:
/cc @kit-ty-kate

Git used into CI still uses INDEX file in version 2. Currently, this version is not properly tested when the current version of the INDEX file is 4. However, to avoid trouble in CI, I did a kind of fix.


dinosaure:
No time to waiting a good alignment of planets to have the sacred fire to green.

PR #382, opened 2020-03-21
use mirage-crypto instead of nocrypto

hannesm:
since mirage-crypto requires OCaml 4.07.0, raise these here as well. also, get rid of custom reseeding code and Mirage_random_test.

### DIFF
--- a/git-http.opam
+++ b/git-http.opam
@@ -15,7 +15,7 @@ build: [
 ]
 
 depends: [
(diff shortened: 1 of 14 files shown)


hannesm:
CI is green, the revdeps run is slightly pointless since it install plotkicadsch which forces various downgrades, and later old tls and x509 is used. but it actually results in all checkmarks :)

dinosaure:
Thanks!

PR #383, opened 2020-03-30
Update README.md with the current version of ocaml-git

dinosaure:


README.md
```diff
@@ -1,109 +1,142 @@
...
-All the objects share a consistent API, and convenience functions are
-provided to manipulate the different objects. For instance, it is
-possible to make a pack file position independent (as the Zlib
-compression might change the relative offsets between the packed
-objects), to generate pack indexes from pack files, or to expand
-the filesystem of a given commit.
+The library comes with some command-line tools called `ogit-*` as a
+Proof-of-concept of the core library which shares a similar interface with
+`git`, but where all operations are mapped to the API exposed `ocaml-git` (and
+hence using only OCaml code). However, these tools are not done to be use. They
```
shonfeder: Just a grammar suggestion since this line was changed, even though you hadn't changed this phrase:

```suggestion
hence using only OCaml code). However, these tools are not meant to be used. They
```

README.md
```diff
@@ -1,109 +1,142 @@
...
 
-All the objects share a consistent API, and convenience functions are
-provided to manipulate the different objects. For instance, it is
-possible to make a pack file position independent (as the Zlib
-compression might change the relative offsets between the packed
-objects), to generate pack indexes from pack files, or to expand
-the filesystem of a given commit.
+The library comes with some command-line tools called `ogit-*` as a
+Proof-of-concept of the core library which shares a similar interface with
+`git`, but where all operations are mapped to the API exposed `ocaml-git` (and
```
shonfeder:
```suggestion
`git`, but where all operations are mapped to the API exposed by `ocaml-git` (and
```

README.md
```diff
@@ -1,109 +1,142 @@
...
-```
+```sh
 $ opam install git
+$ opam install git-unix
 ```
 
+As a MirageOS project, `ocaml-git` is system agnostic. However, it provides a
+`git-unix` package which uses UNIX _syscall_ and it able to introspect a Git
+repository as you usually know. However, `ocaml-git` handles only Git objects
+and it does not _populate_ your filesystem as `git` does.
```
shonfeder: This is a super helpful clarification. I'd even suggest moving this caveat up to the top of the readme, as it makes it very clear that this library is not suitable for many use cases which a user might initial expect from a library called "git".

PR #384, opened 2020-04-28
Fix build for angstrom 0.14.0

sternenseemann:
Add `~consume:Prefix` argument to `parse_string` and `parse_bigstring`.
This achieves the old behavior of the two functions.


sternenseemann:
`angstrom` 0.14.0 hasn't been released to opam yet, so this PR will become relevant in a couple of days, I suppose.

sternenseemann:
angstrom 0.14.0 is out on OPAM now, might be worth to re run the CI.

dinosaure:
I think we can use `All` in these contexts where we should parse all the given input.

sternenseemann:
Rebased to change that. Was a bit worried this might cause problems because of your comment in `helper.ml:110`, but tests seem to run fine.

dinosaure:
> Was a bit worried this might cause problems because of your comment in helper.ml:110, but tests seem to run fine.

Yes, it's another issue but it should be fine, thanks. I will prepare a release of `encore`, merge this PR and prepare a release of `ocaml-git`, then, thanks!

sternenseemann:
Awesome :)

dinosaure:
Sorry for the delay.

Issue #391, opened 2020-06-08
Bug with INDEX file


dinosaure:
See #390 

It seems that `ocaml-git` has some trouble with a specific INDEX file:
```
+1010173us git.index  [DEBUG] Reading 0 byte(s) of the file-descriptor
+1010190us git.index.decoder [DEBUG] We has 0 byte(s), we expect 1 byte(s) to hash, we expect 0 byte(s) to extension.
```

The version of Git used is:
```
# git --version
git version 2.25.4
```

(reportig by @talex5)

talex5:
Note that this happened while running ocaml-git's unit tests. If you can detect when this has gone wrong (is 0 bytes a legal value here?) and crash, instead of writing GBs of data to the log endlessly, that would be better for the CI.

dinosaure:
The CI is fine about `git-index` now so I can close this PR.

PR #393, opened 2020-06-29
Encore

dinosaure:


dinosaure:
AppVeyor fails because `encore` was not available into the repository but it's ok to merge. `index` tests was discarded due to an infinite loop (see #391) but I will fix it at the next big release.

PR #398, opened 2020-09-16
implement ogit-http-fetch

ulugbekna:
- [ ] fix incorrect order of cli arguments to the program
- [ ] write some tests

### DIFF
--- a/git-cohttp-mirage.opam
+++ b/git-cohttp-mirage.opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer:   [ "thomas@gazagnaire.org"
+                "romain.calascibetta@gmail.com" ]
+authors:      "Thomas Gazagnaire"
+license:      "ISC"
+homepage:     "https://github.com/mirage/ocaml-git"
+bug-reports:  "https://github.com/mirage/ocaml-git/issues"
+dev-repo:     "git+https://github.com/mirage/ocaml-git.git"
+doc:          "https://mirage.github.io/ocaml-git/"
+synopsis:     "A package to use HTTP-based ocaml-git with MirageOS backend"
+
+build: [
+  ["dune" "subst"]
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j1" "--no-buffer"] {with-test}
+]
+
+depends: [
+  "ocaml"             {>= "4.07.0"}
+  "dune"              {>= "2.6.0"}
+  "git"
+  "cohttp-mirage"
+]
+
+pin-depends: [
+  [ "conduit-mirage.dev" "git+https://github.com/dinosaure/ocaml-conduit.git#9815aa091e9c285e236e1deffb0dc48723ac72a0" ]
+  [ "cohttp.dev" "git+https://github.com/dinosaure/ocaml-cohttp.git#9f19f9b8216f2546f2917b08ff227be5630858bf" ]
+  [ "cohttp-lwt.dev" "git+https://github.com/dinosaure/ocaml-cohttp.git#9f19f9b8216f2546f2917b08ff227be5630858bf" ]
+  [ "cohttp-mirage.dev" "git+https://github.com/dinosaure/ocaml-cohttp.git#9f19f9b8216f2546f2917b08ff227be5630858bf" ]
+]
--- a/git-cohttp-unix.opam
+++ b/git-cohttp-unix.opam
@@ -1,0 +1,30 @@
+opam-version: "2.0"
+maintainer:   [ "thomas@gazagnaire.org"
+                "romain.calascibetta@gmail.com" ]
+authors:      "Thomas Gazagnaire"
+license:      "ISC"
+homepage:     "https://github.com/mirage/ocaml-git"
+bug-reports:  "https://github.com/mirage/ocaml-git/issues"
+dev-repo:     "git+https://github.com/mirage/ocaml-git.git"
+doc:          "https://mirage.github.io/ocaml-git/"
+synopsis:     "A package to use HTTP-based ocaml-git with Unix backend"
+
+build: [
+  ["dune" "subst"]
+  ["dune" "build" "-p" name "-j" jobs]
+  ["dune" "runtest" "-p" name "-j1" "--no-buffer"] {with-test}
+]
+
+depends: [
+  "ocaml"             {>= "4.07.0"}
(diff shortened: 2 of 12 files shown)


src/git-unix/ogit-http-fetch/main.ml
```diff
@@ -0,0 +1,169 @@
...
+       when it is attached to a terminal, unless -q is specified. This flag \
+       forces progress status even if the standard error stream is not \
+       directed to a terminal."
+    in
+    Arg.(value & flag & info [ "progress" ] ~doc)
+
+  let uri =
+    let parse str = Ok (Uri.of_string str) in
+    let print = Uri.pp_hum in
+    Arg.conv ~docv:"<uri>" (parse, print)
```
dinosaure, line 115: I think, instead to use `Uri.of_string` in this argument, you can use `Smart_git.endpoint_of_string` directly here. By this way, `repository` will be a correct value in `main`.

However, if we go to this way, we extend the ability of `Sync` to `fetch`/`clone` over:
- HTTP / HTTPS
- SSH
- TCP

And `ogit-http-fetch` will be more `ogit-fetch` then. To handle these protocols, you should fill `resolvers` with an SSH implementation and an other TCP implementation on the 9418 port (the port used when we clone with `git://`).

src/git-unix/ogit-http-fetch/main.ml
```diff
@@ -0,0 +1,169 @@
+module Store = Git_unix.Store
+module Sync = Git_unix.Sync (Store) (Git_curl)
```
dinosaure, line 2: May be it's time to make a simple `Git_cohttp` with `cohttp-lwt-unix` and use it instead `git-unix.curl`. An example of what we should have is available inside `src/git-mirage/` (but this one use `Cohttp_mirage`).

src/git-unix/ogit-http-fetch/main.ml
```diff
@@ -0,0 +1,169 @@
...
+    (directory : string option) (repository : Uri.t) =
+  let ( >>?= ) = Lwt_result.bind in
+  let ( >>!= ) v f = Lwt_result.map_err f v in
+  let root : Fpath.t =
+    Fpath.v @@ match directory with Some p -> p | None -> Sys.getcwd ()
+  in
+  let resolvers : Conduit.resolvers =
+    Conduit_lwt.empty
+    |> Conduit_lwt.add Conduit_lwt.TCP.protocol
+         (Conduit_lwt.TCP.resolve ~port:80)
```
dinosaure, line 72: Into details and if we use `Git_curl`, the HTTP implementation does not use `resolvers` at all:

https://github.com/mirage/ocaml-git/blob/master/src/git-curl/git_curl.ml#L9-L18

However, if we start to use `cohttp-lwt-unix`, we must use:
- `Cohttp_lwt_unix.Net.empty` (which handles TCP connection on `*:80` and `*:443`)
- Make sure to use a resolver which can handle HTTP and HTTPS connection

src/git-unix/ogit-http-fetch/main.ml
```diff
@@ -0,0 +1,169 @@
...
+      & info [ "output" ] ~doc ~docv:"<output>")
+
+  let progress =
+    let doc =
+      "Progress status is reported on the standard error stream by default \
+       when it is attached to a terminal, unless -q is specified. This flag \
+       forces progress status even if the standard error stream is not \
+       directed to a terminal."
+    in
+    Arg.(value & flag & info [ "progress" ] ~doc)
```
dinosaure, line 110: Unfortunately, from what I did about `Smart_git`, it's not possible (but should be easy to extend the API) to print what the server does. Indeed, when we do a `fetch`, the server is able to _write_ something to the terminal such as:
```
remote: Enumerating objects: 193, done.
remote: Total 193 (delta 0), reused 0 (delta 0), pack-reused 193
```

Such outputs is available here:
https://github.com/mirage/ocaml-git/blob/436e84defab9b41c6024d5462584e27fa3fb5f33/src/not-so-smart/protocol.ml#L549-L558

But I just decided to ignore them:

https://github.com/mirage/ocaml-git/blob/436e84defab9b41c6024d5462584e27fa3fb5f33/src/not-so-smart/fetch.ml#L118-L119

 As a good introspection of `Smart_git` it can be good to extend the API to be able to print such message.

src/git-unix/ogit-http-fetch/main.ml
```diff
@@ -0,0 +1,169 @@
...
+    Conduit_lwt.empty
+    |> Conduit_lwt.add Conduit_lwt.TCP.protocol
+         (Conduit_lwt.TCP.resolve ~port:80)
+  in
+  let remote_references = List.map fst references in
+  Lwt.return @@ Smart_git.endpoint_of_string (Uri.to_string repository)
+  >>!= store_err
+  >>?= fun endpoint ->
+  Store.v root >>!= store_err >>?= fun store ->
+  Sync.fetch ~resolvers endpoint store (`Some remote_references) >>!= sync_err
```
dinosaure, line 79: The current behavior with `Git_unix.Sync` is to update a local reference A to what the same reference points out into the server. Currently, it's not _exactly_ what `git fetch` does. In fact, with `git fetch`:
- if you say `A:B`, you will try to get the reference A (which exist into the server) but you will set then locally the reference B
- if you say `A`, this is currently the behavior of `Git_unix.Sync.fetch`
- if you say `A:`, you will save the hash into `remotes/A` (and into `FETCH_HEAD`)
- and `:B` is wrong about fetching if I remember correctly (or, at least, require a more complex context)

It should be good to have the same behavior here - and may be update `Git_unix.Sync` to do properly these operations.

src/git-unix/ogit-http-fetch/main.ml
dinosaure: NOTE: at this level, it's fine to require `awa-ssh`/`cohttp-lwt-unix` and a TLS stack to be able to handle HTTPS and SSH.

ulugbekna:
I force-pushed implementation of `git-cohttp-mirage` and `git-cohttp-unix` based on generic `git-cohttp`. 

I could create another PR for these changes (as they are not directly related to ogit-fetch), but my work on ogit-fetch would still need to be based on these changes, so why not to stack them up. 

Let me know if you would prefer a separate PR :-)

dinosaure:
I think, for this specific task, it should better to have an other PR yes :+1:.

PR #400, opened 2020-09-17
Implements a generic `git-cohttp` with more specific `git-cohttp-unix` and `git-cohttp-mirage` based on it

ulugbekna:


ulugbekna:
I would especially appreciate feedback on opam file dependencies (however, if they have some problem, the CI should discover it right?)

dinosaure:
Yes, `cohttp` and `conduit` has several updates and `pin-depends` point to old version of them (old commit), I re-updated them to use last version of `cohttp` and `conduit` - it's a bit annoying at this stage but we should be close to a release of `conduit`.

ulugbekna:
Sounds good, thanks! 

Feel free to merge or let me know if there are things that can be improved :)

dinosaure:
So it seems that `ocamlgraph` is not yet available from OPAM however, OCaml-CI told to us that all is fine. Ready to merge.

PR #402, opened 2020-09-23
Add shallow module as a possible implementation of Git's shallows

dinosaure:


dinosaure:
At this stage, we are able to `clone` with `--depth=1` and we gracefully save shallowed commits. I need to add more tests now.

PR #404, opened 2020-09-28
Fix Sync.fetch behavior to save not only local refs

ulugbekna:
- moves repetitive fetching from `Git_unix.fetch` and `Mem.fetch` to `Sync.fetch`
- changes `Sync.fetch` behavior to save fetched ref `src` either as `dst` ref if provided by the user or as `src`

### DIFF
--- a/src/git/mem.ml
+++ b/src/git/mem.ml
@@ -458,14 +458,9 @@ struct
     let map _ _ ~pos:_ _ = assert false
   end
 
-  open Lwt.Infix
-
   include Sync.Make (Git_store.Hash) (Cstruct_append) (Index) (Conduit)
             (Git_store)
             (HTTP)
-
-  let ( >>? ) x f =
-    x >>= function Ok x -> f x | Error err -> Lwt.return_error err
 
   let stream_of_cstruct ?(chunk = 0x1000) payload =
     let stream, emitter = Lwt_stream.create () in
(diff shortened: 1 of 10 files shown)


dinosaure:
I merged #405 before this PR which introduces conflicts, I can take my time on this part if you want. Otherwise, the PR seems good.

ulugbekna:
I'll rebase a little later. Want to check something first.

ulugbekna:
Since this is user-facing code, I think it's important to get this right:
ogit fetch behavior still deviates from what real git does: git saves fetched branches into `FETCH_HEAD` and/or `<repo_name>/<refname>`, while we directly save to `<refname>` because 

  1) I think we don't have FETCH_HEAD
  2) we don't support `remote`s as in `git add remote origin <URL>`, which would allow to save src refname `master` in `origin/master` instead of overwriting the current master

It seems like the previous ogit fetch did the same.

dinosaure (approved):

dinosaure:
May be it will be useful to add tests. In `test/smart`, we can make a new `alcotest` executable which apply `Git.Mem.Sync.fetch` with a special `conduit`'s flow which send an empty PACK file. The idea then is to see if references are set as expected.

src/git/mem.ml
```diff
@@ -493,30 +488,16 @@ struct
     let index = Carton.Dec.Idx.Device.create t_idx in
     let src = Cstruct_append.key t_pck in
     let dst = Cstruct_append.key t_pck in
+    let create_idx_stream () =
```
dinosaure, line 491: This is exactly what I thought to do :+1:.

test/smart/test.ml
```diff
@@ -232,13 +231,115 @@ let empty_repository_fetch =
...
+        | `Store err -> Alcotest.failf "%a" Git.Mem.Store.pp_error err
+        | `Sync err -> Alcotest.failf "%a" Sync.pp_error err
+        | `Bad_input err -> Alcotest.failf "%a" R.pp_msg err)
+  >|= ignore
+
+(** test Smart_git-related functionality: *)
+
+(* XXX(dinosaure): [tmp] without systemic deletion of directories. *)
+
+module SGit =
```
dinosaure, line 302: I'm not a big fan of this renaming.

test/smart/test.ml
```diff
@@ -31,30 +28,67 @@ let () = Fmt_tty.setup_std_outputs ~style_renderer:`Ansi_tty ~utf_8:true ()
...
+
+let ref_contents =
+  Alcotest.testable
+    (Git.Reference.pp_contents ~pp:Uid.pp)
+    (Git.Reference.equal_contents ~equal:Uid.equal)
+
+(** tmp dir management: *)
+
+(** to keep track of directories created by unit tests and clean them up afterwards *)
+module LocalTmps = struct
```
dinosaure, line 66: We usually use `snake_case` style instead to use `CamlCase` for a module name, can you replace it by `Local_temp`?

dinosaure:
Thanks for this work :+1:

PR #405, opened 2020-09-28
add progress reporting capability to fetch

ulugbekna:
Adds progress reporting capability (as arguments) to fetch through all fetch "levels":

`Nss.Fetch.fetch_v1 <- Smart_git.fetch <- Sync.fetch <- Git_unix/Mem.fetch`

### DIFF
--- a/src/git-unix/git_unix.ml
+++ b/src/git-unix/git_unix.ml
@@ -718,14 +718,15 @@ module Sync (Git_store : Git.S) (HTTP : Smart_git.HTTP) = struct
     Lwt.async fill;
     fun () -> Lwt_stream.get stream
 
-  let fetch ~resolvers edn store ?version ?capabilities want =
+  let fetch ?(push_stdout = ignore) ?(push_stderr = ignore) ~resolvers edn store
+      ?version ?capabilities want =
     let dotgit = Git_store.dotgit store in
     let temp = Fpath.(dotgit / "tmp") in
     tmp temp "pack-%s.pack" >>= fun src ->
     tmp temp "pack-%s.pack" >>= fun dst ->
     tmp temp "pack-%s.idx" >>= fun idx ->
-    fetch ~resolvers edn store ?version ?capabilities want ~src ~dst ~idx temp
-      temp
+    fetch ~push_stdout ~push_stderr ~resolvers edn store ?version ?capabilities
+      want ~src ~dst ~idx temp temp
     >>? function
     | `Empty -> Lwt.return_ok None
     | `Pack (hash, refs) ->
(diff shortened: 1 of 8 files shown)


dinosaure (approved):

dinosaure:
Thanks!

hannesm:
I'm curious why pass through such callbacks, where `logs` is already used, and there could be a log source (enable/disable via the Logs library) to modify the output of log messages?

dinosaure:
I think this choice should be made by the end user - currently such outputs is printed by `git` and we should be able to do the same (and use something like `Fmt.pr`/`Fmt.epr`). Of course, in most of cases, the best is to report that with `Logs`.

hannesm:
I don't quite understand. My question is: why should there be two ways to receive output of the git library -- one using `Logs` (for debug / warnings / errors) -- the other passing through `stdout` and `stderr` arguments. To me, such a API design is quite confusing.

dinosaure:
> I don't quite understand. My question is: why should there be two ways to receive output of the git library -- one using Logs (for debug / warnings / errors) -- the other passing through stdout and stderr arguments. To me, such a API design is quite confusing.

Ah I think, you refer to the old implementation of `ocaml-git` which shows (with `logs`) these outputs. With the new version, [I used `ignore`](https://github.com/mirage/ocaml-git/pull/405/files#diff-e979b4c4c5ebf9ce9e9d0160f408f78eL118-R119) so, these outputs are neither printed out into logs nor printed out with `Fmt.pr`/`Fmt.epr`. This PR wants to expose the ability to choose something like `Logs` or `Fmt.pr`/`Fmt.epr` (into a binary such as `ogit-fetch`).

Currently, [the default is `ignore`](https://github.com/mirage/ocaml-git/pull/405/files#diff-e602d07c34712486d444aaf462ea146aR721) but it was not properly exposed to the `{Git_unix,Mem}.Sync` module.

hannesm:
with this change, I get compilation errors in irmin-git (from @dinosaure's ocaml-git branch):

```OCaml
       Values do not match:
         val fetch :
           resolvers:'a ->
           'b -> 'c -> ?version:'d -> ?capabilities:'e -> 'f -> 'g
       is not included in
         val fetch :
           ?push_stdout:(string -> unit) ->
           ?push_stderr:(string -> unit) ->
           resolvers:Conduit.resolvers ->
           Smart_git.endpoint ->
           G.t ->
           ?version:[> `V1 ] ->
           ?capabilities:Smart.Capability.t list ->
           [ `All | `None | `Some of Git__.Reference.t list ] ->
           ((G.hash * (Git__.Reference.t * G.hash) list) option, error)
           result Lwt.t
       File "src/git/sync.mli", lines 27-36, characters 2-67:
         Expected declaration
       File "src/irmin-git/irmin_git.ml", line 1075, characters 6-11:
         Actual declaration
```

PR #407, opened 2020-09-29
git-cohttp-unix and git-cohttp-mirage use git-cohttp

hannesm:


dinosaure (approved):
Thanks, I did that into #402 but did not extract it yet and make a new PR (this change is needed due to [this commit](https://github.com/mirage/ocaml-cohttp/pull/692/commits/2f346f1aef0312f29e3283e176607d46c972620d))

PR #413, opened 2020-10-07
Fix the ephemeron used to simulate and temporary device for the memory back-end of Git

dinosaure:
Spotted by @ulugbekna, I need to implement some tests.

### DIFF
--- a/src/carton/idx.ml
+++ b/src/carton/idx.ml
@@ -531,6 +531,7 @@ module Device = struct
   let create tbl =
     let key = ref Key in
     Ephemeron.K1.set_key tbl key;
+    Ephemeron.K1.set_data tbl (ref Bigstringaf.empty);
     key
 
   let project tbl uid =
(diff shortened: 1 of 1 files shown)


ulugbekna:
I think this causes 

```
index out of bounds
          Raised by primitive operation at file "src/carton/idx.ml" (inlined), line 43, characters 29-48
          Called from file "src/carton/idx.ml", line 64, characters 10-27
          Called from file "src/git/mem.ml", line 322, characters 6-223
          Called from file "src/git/sync.ml", line 232, characters 8-64
          Called from file "src/core/lwt.ml", line 1867, characters 23-26
```

with payload `empty_repository_fetch` in `smart/test.ml`

ulugbekna:
To note: the same payload doesn't cause problems in git-unix

dinosaure:
After investigation, we got an error about IDX temporary file because we did the _projection_ of it before to do anything on it - I mean, we want to get contents of the file before to do receive and analyze a PACK file. So, the case should not appear - and it's not a bug as is (when such usage should not happen).

However, the patch assert that in any context, we have, at least, an empty contents. It does not change anything for us but, with this patch. the code looks similar to `Cstruct_append` (which is better). Let's start to merge.

PR #418, opened 2020-10-27
Delete pin-depends on OPAM files about conduit

dinosaure:


dinosaure:
Closed by #428 which will be merge as soon as I can.

PR #420, opened 2020-11-27
[WIP] Implement Git Wire Protocol v2 

ulugbekna:
This is a WIP implementation of git wire protocol v2 (spec available [here](https://github.com/git/git/blob/master/Documentation/technical/protocol-v2.txt)).

It's not really ready for a review, but if you still want to see how things are going, I'd recommend going from the oldest commit to the newest. Things should make sense.


src/not-so-smart/decoder.ml
```diff
@@ -124,34 +124,31 @@ let at_least_one_line decoder =
...
-  !code
+(** reads off 4 bytes from [decoder.buffer] starting at [decoder.pos] and interprets read
+    bytes as hex and converts to int.
+    Why unsafe:
+    @raise Invalid_argument if there are no 4 bytes to read, i.e.,
+           [decoder.max - decoder.pos < 4]  *)
+let pkt_len_unsafe (decoder : decoder) =
+  let hex = Bytes.of_string "0x0000" in
+  Bytes.blit decoder.buffer decoder.pos hex 2 4;
+  Bytes.to_string hex |> int_of_string
```
dinosaure, line 139:
```ocaml
int_of_string (Bytes.to_string hex)
```

is better and you can safely use `Bytes.unsafe_to_string` in that case to avoid a double-allocation.

src/not-so-smart/decoder.ml
```diff
@@ -184,12 +188,10 @@ let get_pkt_len decoder =
    the protocol error to another layer (eg. [carton] when it received finally a
    __not-full__ PACK file). The goal is to be more resilient at this layer. *)
 
-let error_end_of_input decoder () = fail decoder `End_of_input
-
 let reliable_pkt k decoder () =
   match get_pkt_len decoder with
   | Some _len ->
-      let hdr = Fmt.strf "%04X" (decoder.max - decoder.pos) in
+      let hdr = Fmt.str "%04X" (decoder.max - decoder.pos) in
```
dinosaure, line 194: It should be a part of 09a28e4

src/not-so-smart/decoder.ml
```diff
@@ -273,13 +268,13 @@ let peek_while_eol decoder =
...
-      not (!chr == '\n' && !has_cr) )
+      not (!chr = '\n' && !has_cr) )
   do
-    has_cr := !chr == '\r';
+    has_cr := !chr = '\r';
     incr idx
   done;
 
-  if !idx < end_of_input decoder && !chr == '\n' && !has_cr then (
+  if !idx < end_of_input decoder && !chr = '\n' && !has_cr then (
```
dinosaure, line 277: Can you split this update out of this PR (and make a new PR just about that)?

dinosaure:
So the PR seems fine as is, I need to go to some details but these commits should be proposed into another PR:
09a28e4
f24b46e & c65712c

src/not-so-smart/decoder.ml
ulugbekna: just to be clear:
1. you don't want me to use `|>`? 
2. from your message it seems like `|>` prevents us from using `unsafe_to_string`. If that's true, why is it so?

src/not-so-smart/protocol.ml
```diff
@@ -296,10 +296,6 @@ module Decoder = struct
         Fmt.pf ppf "Invalid result command (%S)" raw
     | `Unexpected_flush -> Fmt.string ppf "Unexpected flush"
 
-  let rec prompt_pkt ?strict k decoder =
-    if at_least_one_pkt decoder then k decoder
-    else prompt ?strict (prompt_pkt ?strict k) decoder
-
```
dinosaure, line 302: Not sure about such deletion, I know that usually, I re-bind an OCaml value with the same name several such as:
```ocaml
let prompt_pkt = ...
let prompt_pkt = ... prompt_pkt ...
let prompt_pkt = ... prompt_pkt ...
...
```

So, in my mind, such deletion will disturb behaviors of functions defined then. Why you deleted this code?

dinosaure:
The pull-request has some style updates which should be out of that. I don't have strong opinion about style but they should be done with some others PRs (for example, I saw some update about `push.ml`). Is it possible to split again the PR with style and the impl. of the protocol v2?

src/not-so-smart/protocol.ml
ulugbekna: I moved `prompt_pkt` from `protocol.ml` to `decoder.ml` because 

1. `protocol.ml` contains `Decoder` module, which implements decoding for values of the pack protocol. But `prompt_pkt` is a function that is _not_ specific to protocol and could be defined on a lower and more general layer such as `decoder.ml`.
2. I also needed it in `proto_vals_v2.ml`, so it made sense to move this general and low-level function to `decoder.ml` which specifically deals with pkt-line decoding
3. Moving `prompt_pkt` to `decoder.ml` should be safe because all references to `prompt_pkt` in `protocol.ml` still refer to that value but "imported" from `decoder.ml`.

PR #422, opened 2020-12-09
replace '==' with '=' in 'nss/decoder' and 'ogit-fetch'

ulugbekna:
Changes are in modules that I am working on now. Left `==` uses in `carton` et al. unchanged.


dinosaure (approved):
Thanks, I will merge it when the CI says that is good.

PR #423, opened 2020-12-09
Replace 'prelude' with 'uses_git_transport'

ulugbekna:
Previously, we called a prelude the message sent from the client to server requesting a service. Such a request message is necessary only for `git://` transport, so the code should be more understandable if we give it a more explicit name. 

A note: 

@dinosaure 's comment in the code

```ocaml
(* XXX(dinosaure): [prelude] is the only tweak needed between git:// and SSH. *)
```

in reality means that `prelude` is the only tweak needed between `git://` and (SSH and `file://`), but since we don't support `file://`, the comment is true. Also note that `http://` is quite different from all other transports and is not considered in this context of initial request of service from server.

### DIFF
--- a/src/not-so-smart/default.ml
+++ b/src/not-so-smart/default.ml
@@ -1,3 +1,8 @@
+(** default[1] negotiator implementation
+
+   [1] "default" as defined in the canonical git implementation in C,
+       see https://github.com/git/git/tree/master/negotiator *)
+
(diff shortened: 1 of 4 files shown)


dinosaure (approved):
Thanks :+1: and you are true, the code seems more understable. Waiting the CI and ready to merge.

PR #426, opened 2020-12-16
Upgrade ocamlformat to 0.16.0

ulugbekna:
Since ocamlformat 0.16.0 is released already, upgrading to it.

I looked into why spaces between parentheses and their content are missing but couldn't find a related change in the ocamlformat [changelog](https://github.com/ocaml-ppx/ocamlformat/blob/master/CHANGES.md). I can ping someone from the ocamlformat team if that's important to keep. 

#419 could be closed.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version=0.15.0
+version=0.16.0
 module-item-spacing=compact
 break-struct=natural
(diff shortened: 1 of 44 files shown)


dinosaure:
Considering 1ff8f22 in #427, this PR can be closed.

PR #429, opened 2021-01-07
Delete the useless verify function and start an SSH connection with a pre-condition over the given Smart_git.Endpoint.t

dinosaure:
No regression here, I tested the unikernel locally and even if the user gives to use an SSH key with a GRI remote, we start a GRI (with the prelude) connection. /cc @ulugbekna


dinosaure:
Fix an issue highlghted by @ulugbekna here: https://github.com/mirage/ocaml-git/pull/428#issuecomment-756027461

ulugbekna:
Great, thanks!

PR #436, opened 2021-01-12
Handle IPv4 when the user gives git@[IPv4]:path as the remote

dinosaure:
A minor patch (followed by a release) to be able to gives to an unikernel something like: `git@[10.0.0.1]:path` (mostly when we want to start a primary DNS server.


dinosaure:
Note that even if it break API for `git.nss`, nothing need to be updated on the Irmin side.

PR #439, opened 2021-01-16
git-unix: make index and regression test work without global gitconfig

sternenseemann:
The tests of git-unix fail if there's no global `.gitconfig` setting
user.name and user.email in every `git commit` being run. This easily
fixed by setting a local git config per repository/store used.

This fixes the tests failing in sandboxed environments or on systems
without git set up properly.

### DIFF
--- a/test/index/test.ml
+++ b/test/index/test.ml
@@ -183,6 +183,11 @@ let delete_should_be_empty =
   let run path =
     let open Rresult in
     Bos.OS.Dir.with_current path @@ fun () ->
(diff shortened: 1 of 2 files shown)


dinosaure:
Thanks and good catch! Can you run an `dune build @fmt --auto-promote` to make the CI happy?

sternenseemann:
Done :) One ocaml-ci fails, but I'm not sure how to fix that warning or whether I should?!

dinosaure:
I will fix the error on an other side, I think it's ready to merge, thanks!

dinosaure:
I will fix the error on an other side, I think it's ready to merge, thanks!

PR #447, opened 2021-01-28
Remove unused Lwt_io module

dinosaure:
/cc @ulugbekna


ulugbekna:
We could also remove `carton/higher.ml`, which is also unused.

ulugbekna:
Thanks 👍😊

PR #448, opened 2021-01-28
Drop the support of 4.07.0 and remove stdlib-shims

dinosaure:
`bigarray-compat` still exists for:
- `bigstringaf`
- `decompress`

Which are a long-term support of OCaml versions.

### DIFF
--- a/src/carton/dune
+++ b/src/carton/dune
@@ -2,12 +2,12 @@
  (name carton)
  (modules carton dec enc h idx sigs zh)
  (public_name carton)
- (libraries stdlib-shims ke duff bigarray-compat optint checkseum
-   decompress.de decompress.zl bigstringaf psq fmt))
+ (libraries ke duff optint checkseum bigarray decompress.de decompress.zl
+   bigstringaf bigarray-compat psq fmt))
 
 (library
  (name thin)
  (modules thin)
  (public_name carton.thin)
- (libraries optint checkseum decompress.de decompress.zl bigarray-compat
-   bigstringaf logs carton cstruct ke))
+ (libraries bigarray-compat optint checkseum decompress.de decompress.zl
+   bigarray bigstringaf logs carton cstruct ke))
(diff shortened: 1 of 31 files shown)


src/not-so-smart/dune
```diff
@@ -8,8 +8,7 @@
  (name smart)
  (public_name git.nss.smart)
  (modules smart filter capability state protocol)
- (libraries git.nss.pkt-line stdlib-shims result rresult ipaddr domain-name
-   astring fmt))
+ (libraries git.nss.pkt-line result rresult ipaddr domain-name astring fmt))
```
hannesm, line 11: it's likely that result can as well be dropped here (result is compat with < 4.03)

src/carton/dune
```diff
@@ -2,12 +2,12 @@
  (name carton)
  (modules carton dec enc h idx sigs zh)
  (public_name carton)
- (libraries stdlib-shims ke duff bigarray-compat optint checkseum
-   decompress.de decompress.zl bigstringaf psq fmt))
+ (libraries ke duff optint checkseum bigarray decompress.de decompress.zl
```
hannesm, line 5: I'm not sure why bigarray is needed here now..

src/carton/dune
```diff
@@ -2,12 +2,12 @@
...
+   bigstringaf bigarray-compat psq fmt))
 
 (library
  (name thin)
  (modules thin)
  (public_name carton.thin)
- (libraries optint checkseum decompress.de decompress.zl bigarray-compat
-   bigstringaf logs carton cstruct ke))
+ (libraries bigarray-compat optint checkseum decompress.de decompress.zl
+   bigarray bigstringaf logs carton cstruct ke))
```
hannesm, line 13: same as above, I think bigarray can be dropped

src/not-so-smart/dune
dinosaure: Unfortunately, it seems that `rresult` depends to `result`. We chosen to use `(implicit_transitive_deps false)` in this project to be sure to be explicit about all dependencies. So if I remove `result`, `dune` complains that `Rresult.t` is not compatible with `Stdlib.result`.

For more information, you can see this issue: https://github.com/mirage/irmin/pull/1024

PR #450, opened 2021-01-28
Fix a bug about the missed minimal len when we read the queue on the unixiz.ml

dinosaure:


dinosaure:
Hard to reproduce the bug but I got the situation when I tried to fetch a Git repository with SSH.

PR #454, opened 2021-01-30
Unmonad mmap

dinosaure:
If we assume that `mmap` does not block the execution, we can unwrap it from the IO monad. Such update unlock the ability to really take the opportunity to unpack Git objects in concurrency/with true parallelism as Git does. The diff is huge because we update the interface of `carton`.

However, tests are green about decode/encode PACK file and we should don't have errors on large PACK file (however, I need to test such case). Finally, for LWT/Git, we are able to use `Lwt_list.iter_p` and `Lwt_list.map_p` which should be very good when we clone large PACK file.

About tests, a derivation of [minifiber] was done with `Thread` to test, as `git` does, `git index-pack`/`git verify-pack` with real threads (and, in this context, if `carton` works properly into a concurrent context). We should know easily if we have a performance improvement, but naively, I can say yes when multiples threads can unpack several objects cooperatively now.

### DIFF
--- a/bench/bench_pack.ml
+++ b/bench/bench_pack.ml
@@ -16,8 +16,7 @@ let map fd ~pos len =
   let res =
     Unix.map_file fd ~pos Bigarray.char Bigarray.c_layout false [| len |]
   in
-  let res = Bigarray.array1_of_genarray res in
-  inj res
+  Bigarray.array1_of_genarray res
 
 let fd = Unix.openfile bomb_pack Unix.[ O_RDONLY ] 0o644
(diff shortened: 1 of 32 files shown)


dinosaure:
I added a simple a tool which is a pure OCaml implementation of `git-verify-pack`. It permits to us to see if `carton` still is able to handle large PACK files as the one from `ocaml/ocaml`. We are able to do macro-benchmarking on this tool to compare performance.

Indeed, this part of the analyze is the bigger one when we want to fetch/clone.

PR #456, opened 2021-02-01
Fix advertised_refs when we want to push first.

dinosaure:


dinosaure:
It's a bit hard to reproduce the bug, two things appears:
- `shallows` does not appear after `advertised-refs` (without references)
- a push without side-band leads to a possible error (which is not the case with Irmin)
- Fixed by #455, if the reference does not exists, we replace it by `Create` (instead `Update`) and avoid the `Not_found` error

However, it seems that Git complains if we want to push several time the "same" commit (not the same date). I try to check PACK file and GIt objects and they are well-formed (even if Git complains), I think the error is mostly about a _force-push_ but we should investigate that at one point.

PR #457, opened 2021-02-02
Adapt to conduit 2.3.0

samoht:
Follow-up of https://github.com/mirage/ocaml-conduit/pull/376 and https://github.com/mirage/ocaml-cohttp/pull/742

### DIFF
--- a/git-cohttp-mirage.opam
+++ b/git-cohttp-mirage.opam
@@ -19,6 +19,8 @@ depends: [
   "result" {>= "1.5"}
   "rresult" {>= "0.6.0"}
   "uri" {>= "4.0.0"}
+  "mirage-clock"
+  "x509"
   "alcotest" {>= "1.2.3" & with-test}
   "alcotest-lwt" {>= "1.2.3" & with-test}
   "bigstringaf" {>= "0.7.0" & with-test}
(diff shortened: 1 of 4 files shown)


dinosaure:
A side-effect of that will be about `irmin-mirage-git` which requires `git-cohttp-mirage`. Currently, by the fact that `Cohttp_mirage` is not a functor, transitively, `Git_cohttp_mirage` is not a functor. With this PR, we extend the functor of `irmin-git-mirage` with a `Resolver.S` and `Conduit.S`: [an example when we need `Git_cohttp_mirage`](https://github.com/mirage/irmin/blob/ad2450c4da50a9fe5dab5d074a33cb3bd19511aa/src/irmin-mirage/git/irmin_mirage_git.ml#L61)

I'm still no convince that it's the right solution when we mostly re-functorize what it should be de-functorized by Conduit. Even if `functoria` can (and should) help us on this side, it breaks the complete path between `irmin` and `conduit` (including `cohttp` and `git`).

However, I don't want to block anything so if you think that we should go deeper on this way, I'm ok with this PR - but I will waiting a proper a release of `conduit`/`cohttp` before to merge this PR.

PR #463, opened 2021-02-10
git-unix: Fix new smart test without a global git config

sternenseemann:
The issue fixed here is the same as #439. We also make sure the git
config is set in make_one_commit in case tests are run in an environment
without a global git config.

### DIFF
--- a/test/smart/test.ml
+++ b/test/smart/test.ml
@@ -1620,6 +1620,11 @@ let test_partial_fetch_ssh () =
 
 let make_one_commit path =
   Bos.OS.Dir.with_current path @@ fun () ->
+  Bos.OS.Cmd.run Bos.Cmd.(v "git" % "config" % "user.name" % "test")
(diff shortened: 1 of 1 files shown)


dinosaure:
Ah sorry 👍 I forget about NixOS stuffs, thanks! I will apply `ocamlformat` and merge it.

sternenseemann:
Forgot `ocamlformat` again, force pushed with formatting :)

I was thinking that if I find the time or the next time this situation arises I could probably just refactor the tests by writing a tiny wrapper around "git commit" which ensures the config is always set.

dinosaure:
> Forgot `ocamlformat` again, force pushed with formatting :)
> 
> I was thinking that if I find the time or the next time this situation arises I could probably just refactor the tests by writing a tiny wrapper around "git commit" which ensures the config is always set.

It will be useful indeed :+1:. Thanks.

PR #465, opened 2021-02-10
ref-find-common

ulugbekna:
- make semantically close things closer in file
	- removed underscores from variable names because they usually
	  mean values are not used, which is not the case here
	- use a first-class module to remove rebinding scheduler
	  functionality again and again

### DIFF
--- a/src/not-so-smart/find_common.ml
+++ b/src/not-so-smart/find_common.ml
@@ -1,30 +1,7 @@
 open Sigs
-open Smart_flow
-
-let ( <.> ) f g x = f (g x)
 
 module Log = (val let src = Logs.Src.create "find-common" in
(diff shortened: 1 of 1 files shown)


test/smart/dune
```diff
@@ -16,9 +16,7 @@
 
 (rule
  (alias runtest)
- (package git-unix)
- (enabled_if
-  (= %{os_type} "Unix"))
+ (package git)
```
dinosaure, line 19: Currently these tests can not work on Windows due to named pipes. We don't have any equivalent with that so, to avoid an error on AppVeyor we execute them only for `unix`.

dinosaure:
I'm not really convince by the first class module but let's merge.

PR #468, opened 2021-02-16
Pass client-side capabilities instead server

dinosaure:


dinosaure:
It's a bug about capabilities/version of Git/etc. we pass server-side capabilities instead of clients. We should add a small test to avoid any regressions.

ulugbekna:
Woops, my bad!

dinosaure:
> Woops, my bad!

I don't think that it's your fault 👍 I `blame` this part and it seems that it's me!

hannesm:
Thanks for this patch -- as mentioned on IRC, this indeed solves the issue I encountered when attempting to push to a git remote via ssh. I'm curious whether there is other code that mixes the client and server capabilities (I guess `fetch.ml` is the only other source that cares about capabilities)?

dinosaure:
> I'm curious whether there is other code that mixes the client and server capabilities (I guess fetch.ml is the only other source that cares about capabilities)?

Yes, capabilities are send via `find_common.ml:126` and I double-checked that we send the right side of capabilities.

PR #469, opened 2021-02-17
Implementation of HTTP push and fix bug about commit

dinosaure:
This PR has two things:
- fix a bug about commit
  Indeed, we should have an extra-LF between the commit header and the message. I suspect that Git broke something about that and it silently accepts such commit without verification (`git log` deals with that). However, it seems that `git fsck` checks such assumption and when we want to push over HTTP on GitHub, this command is launched and invalidate our PACK file.

  It's a breaking change! Due to the design of `ocaml-git`/`encore`, old commits (without the double LF) **are not** accepted anymore - we are surely more conservative than Git on this aspect... I advise (/cc @hannesm) to make a fake commit (`git commit --allow-empty -m.`) to sanitize, at least, the last commit (and properly do, then, the `git fetch --depth=1`).

  For the story, I think that the first commit of Linux is **not** well-serialized and it's why I decided to accept such format. May be a solution should be about an optional LF but I'm not sure that I'm able to do that and, at the same time, respect the isomorphism. At least, `ocaml-git` should produce valid commits.

- I implemented the HTTP push and we accept a remote such as: `http://user:password@github.com/repo`. I tested the unikernel example with GitHub and I rewrite the `report-status` decoder to be correct with HTTP responses. Regression tests help me to keep the compatibility with `git://` and SSH (big thank to the old version of me). It still is a pain to parse something which is **not** specified. So HTTP push should work but I did not write tests yet.

### DIFF
--- a/git.opam
+++ b/git.opam
@@ -19,6 +19,7 @@ depends: [
   "dune" {>= "2.6.0"}
   "digestif" {>= "0.8.1"}
   "rresult"
+  "base64" {>= "3.0.0"}
   "result"
   "bigstringaf"
   "bigarray-compat"
@@ -45,7 +46,6 @@ depends: [
   "cmdliner" {with-test}
   "base-unix" {with-test}
   "fpath"
-  "base64" {with-test & >= "3.0.0"}
   "mirage-flow" {>= "2.0.1"}
   "domain-name" {>= "0.3.0"}
   "emile" {>= "1.1"}
--- a/src/git/commit.ml
+++ b/src/git/commit.ml
@@ -146,7 +146,7 @@ module Make (Hash : S.HASH) = struct
       <*> binding ~key:"author" user
       <*> binding ~key:"committer" user
       <*> rep0 extra
-      <*> rest
+      <*> (Encore.Bij.char '\x0a' <$> any) *> rest
 
     let format = Encore.Syntax.map Encore.Bij.(compose obj6 commit) t
   end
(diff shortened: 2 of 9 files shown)


dinosaure:
I'm not really happy to merge it as is but, eh we need the HTTP push.

hannesm:
It is not clear to me what the impact about the breaking change you mentioned above - does it mean "all commits done by ocaml-git won't be valid git commits (for ocaml-git) anymore"? I.e. a clean commit with cgit is required on a repository so that ocaml-git can clone it (with depth=1) -- a repository full of ocaml-git commits is not clonable by ocaml-git anymore (since the commits are bad, and need to be rewritten)?

dinosaure:
> It is not clear to me what the impact about the breaking change you mentioned above - does it mean "all commits done by ocaml-git won't be valid git commits (for ocaml-git) anymore"? I.e. a clean commit with cgit is required on a repository so that ocaml-git can clone it (with depth=1) -- a repository full of ocaml-git commits is not clonable by ocaml-git anymore (since the commits are bad, and need to be rewritten)?

Exactly, a sanitizer of all of your commits with `cgit` should works but it seems that `cgit` accepts both formats until you do a `git fsck`.

A hot fix is to make a fake commit with `cgit`, and do the ferch then with `ocaml-git` & `--depth=1`. This situation appears when we use directly `ocaml-git` but it seems that `irmin` put systematically the second LF (https://github.com/mirage/irmin/blob/3f04145b388f1c73ab685ae0ee423b5021fe7c77/src/irmin-git/irmin_git.ml#L394). So the error does not appear with a commit with `irmin` and a trailing LF will appear then.

PR #470, opened 2021-02-18
A git commit can have an empty message

dinosaure:
Fix mirage/irmin#917. A `git fsck` is launched to check some invariants about format of Git objects. Note that `cgit` silently accepts our commits for a long time without errors. This PR has a breaking change about the API. A patch on `irmin` will coming soon to keep the compatibility.


dinosaure:
Ok, the CI is happy, let's merge and cut a release with all of that.

PR #472, opened 2021-02-19
test/unix: fix write test without a git config

sternenseemann:
This is becoming a similar sight, but easily fixed, updates in NixOS should be done soon as well :)

---

git commit fails if user.email and user.name are not set which may be
the case for sandboxed environments for example. This is easily fixed by
setting a local git config for the current directory before commiting.

### DIFF
--- a/test/unix/test.ml
+++ b/test/unix/test.ml
@@ -258,6 +258,11 @@ let empty_commit =
     let open Rresult in
     R.join
     <.> Bos.OS.Dir.with_current path @@ fun () ->
+        Bos.OS.Cmd.run Bos.Cmd.(v "git" % "config" % "user.name" % "test")
+        >>= fun () ->
+        Bos.OS.Cmd.run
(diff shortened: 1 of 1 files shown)


dinosaure:
Shame on me... I'm really sorry because I see me 2 days ago think about that and... I forgot. Thanks for your work. I will try to find a CI about NixOS on the other side.

sternenseemann:
> I will try to find a CI about NixOS on the other side.

Not sure if it's worth the hassle since you'd need to maintain a mapping from ocaml dependencies to NixOS packages for that in some shape or form. You could also catch this class of failures by having the following before a test run, I think:

```
export HOME="$(mktemp -d)"
set -xe XDG_CONFIG_HOME
```

This would make `git` cease to find its configuration triggering these kind of test execution bugs.

Or we go down a different path: I also could create a temporary global gitconfig for the tests in the NixOS build environment or we do it somewhere in the tests?

PR #474, opened 2021-03-04
A fix about reporting status when we push

dinosaure:
The current situation about `report-status` and protocols is a bit a mess. At this stage, I can say that with `Side_band_64k` and `Report_status` capabilities, all protocols works. I can say that `git://` works with any capabilities. However, SSH and HTTP(s) have some problems if we don't use these capabilities (even if we replace `Side_band_64k` by `Side_band`).

Because the protocol is completely under-documented, it's hard for me to understand where is the bug and reproducibility seems hard when the Git protocol involve along the versions.

However, this PR seems to fix problems for our use case (with `irmin`) and pass regression tests (thanks for the old version of me). Unikernels work with that and we are able to fetch/push correctly _without_ any problems.

- [ ] clean the PR

### DIFF
--- a/carton-git.opam
+++ b/carton-git.opam
@@ -24,7 +24,7 @@ depends: [
   "mmap"
   "fmt" {>= "0.8.7"}
   "base-unix"
-  "decompress" {>= "1.2.0"}
+  "decompress" {>= "1.2.0" & < "1.3.0"}
   "astring" {>= "0.8.5"}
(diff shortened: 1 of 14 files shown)


hannesm:
I can confirm that this branch works fine with my caldav unikernel which pushes to a git remote via ssh (and uses irmin (Irmin_mirage_git.KV_RW) in between).

PR #480, opened 2021-03-11
Fix infinite loop when we don't have enough to output the deflated content

dinosaure:


dinosaure:
For some contexts, `decompress` always requires, at least, 2 bytes to compress something. It was explained why in the `decompress` 's implementation - but it was not correctly explained in the documentation. You can see this long comment about that [here](https://github.com/mirage/decompress/blob/aa4a163535eb53cffa17a17c32e2ed13af085abc/lib/de.ml#L2349-L2367).

So according to this comment, we must protect the loop against a specific situation:
- when the encoder asks to flush the output buffer (so the encoder full-filled the output buffer)
- when it remains 1 free byte (which is not enough for the next call to `Zl.Def.encode`)

In that case, the encoder strictly does nothing and stuck on this situation when it expects something bigger than one byte. Before this patch, we consider that one byte is enough to continue. Now, we stop the process and say that it's not enough to continue to encode the flow atomically.

The situation is difficult to reproduce because it depends on what we want to encode. More concretely, we got this bug in Irmin due to a random generation of a large tree. And, depending on the huffman tree generated from the randomly generated large tree, this situation can appear (mostly when we reach the end of the output and we want to encore an Huffman _op-code_ larger than 8 bits... such as a _distance_ _op-code_).

For the rest of the code, I think it's not a problem because we try with a limited output a compression only on this part of `ocaml-git`. The rest of the code (see `carton`) does not try to encode something into a fixed-size buffer. Then, the code is a bit fragile but I hope that we will be able to replace it with a non-stream API of `decompress`.

dinosaure:
And thanks to @Ngoguey42 who helps me to reproduce the bug and see if the patch fixes the problem.

PR #485, opened 2021-03-16
Fix possible stack overflow in Tree.v

zshipko:
This PR fixes a possible stack overflow in `Tree.v` found when working with very wide trees in Irmin

See https://github.com/mirage/irmin/pull/1313


dinosaure:
I added something related to mirage/encore#31, ready to merge. I will try to do a release this week.

PR #488, opened 2021-03-25
Fix some issues on the UNIX back-end

dinosaure:
It includes:
- some bugs about the path
- A real `mimic` resolver (SSH is not handled) for UNIX (DNS, TCP/IP, HTTP and HTTPS)
- issue about internal resources with PACK file

The goal is to fix an usage of `irmin` (thx to @zoltan-o). I think it's time to implements some tests to really check `git-unix`.


dinosaure:
I clean a bit the distribution when `ogit-fetch` is replaced by `guit.fetch` now. Then, I added a cram test to see if all works (as `irmin` should do when it wants to clone/fetch). I extended a bit `Smart_git.Endpoint.t` to allow an user-defined extension of a transmission protocol as it's the case for `guit.fetch` (see `fifo://pipe/` and `test/guit/fetch.t`).

It brokes the whole distribution but it's for something better when we allow know to really test `git-unix` properly (and ensure some expected behaviors required by `irmin`).

dinosaure:
PS: we should add the SSH support via `execve` of SSH - however, we don't have the possibility to test it when we don't have a network access. (/cc @sternenseemann - I use `git init` and `git commit` in my test, I believe that they should be follow by `git config`?).

test/guit/fetch.t
```diff
@@ -0,0 +1,18 @@
+Tests fetch command
+  $ mkdir simple
+  $ mkdir simple-clone
+  $ cd simple
+  $ git init
```
sternenseemann, line 5: Recent versions of `git` display a warning wenn calling `git init` without either giving `-b` (which is not available for older git versions) or setting `init.defaultBranch` in the global git configuration.

```suggestion
  $ git init 2> /dev/null
```

test/guit/fetch.t
```diff
@@ -0,0 +1,18 @@
...
+  Initialized empty Git repository in $TESTCASE_ROOT/simple/.git/
+  $ git commit --allow-empty -q -m .
```
sternenseemann, line 7: Need `user.name` and `user.email` to be set before this.

sternenseemann (changes requested):

sternenseemann:
> PS: we should add the SSH support via `execve` of SSH - however, we don't have the possibility to test it when we don't have a network access.

Sounds good to test nevertheless, should just be skipped if no network access is present :)

test/guit/fetch.t
dinosaure: It seems that `dune runtest` changes my global config (see 6aaa6a4), I will try to find a way to locally set these values.

sternenseemann: `git config` without `--global` sets it per repository (you just have to run it after *every* `git init`.

Maybe it's time for a sandboxed dune test runner ;)

test/guit/fetch.t
```diff
@@ -2,11 +2,11 @@ Tests fetch command
   $ mkdir simple
   $ mkdir simple-clone
   $ cd simple
+  $ git init 2> /dev/null
+  Initialized empty Git repository in $TESTCASE_ROOT/simple/.git/
   $ git config init.defaultBranch master
```
sternenseemann, line 7: That line is not necessary anymore, then?

dinosaure: Hmmhmm, it seems required and even if we do `&> /dev/null`, `git init` in `cram` still prints something...

dinosaure:
Let's merge, I will try to cut a release as soon as I can to fix the problem on the irmin layer then.

PR #499, opened 2021-04-08
Add missing library depends on `fmt` and `logs`

CraigFe:
This project uses `(implicit_transitive_deps false)`, but doesn't
specify all transitive deps in the case of these libraries. With the
main releases, this is OK since they don't build with Dune, but it's a
bug when using the Dune overlays / `opam monorepo` for development.

### DIFF
--- a/src/git-top/dune
+++ b/src/git-top/dune
@@ -8,4 +8,5 @@
   git
   fmt
   logs
+  logs.fmt
   compiler-libs.toplevel))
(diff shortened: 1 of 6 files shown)


dinosaure:
Strange that I did not notice it when the distribution comes with `(implicit_transitive_deps false)` - I think it's a diff between my OPAM environment and the `opam monorepo` repositories when it concerns _non-dunified_ libraries. Thanks.

PR #502, opened 2021-04-13
Use Ns decompress in atomic get and add

clecat:
Use the non-streamable API when the transactions are supposed to be atomic.
It uses this [pull request](https://github.com/mirage/decompress/pull/102) which should be accepted soon


dinosaure:
Waiting the release of `decompress.1.4.0` but it seems good, thanks :+1:.

PR #518, opened 2021-05-13
carton: Add compatibility with OCaml 4.13

kit-ty-kate:
OCaml 4.13 added `Int64.max` and thus made theses two lines break.
In general I would encourage people to avoid opens when possible at makes it more likely to have code that breaks in this fashion.


dinosaure:
Indeed, thanks for this patch!

kit-ty-kate:
Any chance this could be released? The first beta for 4.13 is around the corner.

dinosaure:
Yes, I would like to release `checkseum`, `decompress` and `paf` before 👍

PR #523, opened 2021-07-28
Fix big-endian support

dinosaure:


dinosaure:
- [ ] `checkseum.c` does not work on big-endian machine

PR #537, opened 2021-11-07
drop unneeded mirage-protocols dependency

hannesm:
while investigating reverse dependencies (via https://opam.ocaml.org/packages/mirage-protocols/) I encountered ocaml-git which does not use this package


dinosaure:
Thanks!

PR #539, opened 2021-11-12
Delete http and use eyeballs

### LINKED ISSUE #513: git and irmin: remove hard-wired http dependencies
At the moment, irmin-mirage-git depends on conduit/cohttp/git-paf, so there's no way to configure a irmin+git unikernel which only supports the git-over-tcp (git-over-ssh) protocol. in some use cases, I'd like to avoid the dependency cone of http/ssh/..., and only use the tcp (ssh) transport mechanism.

AFAICT some API changes are necessary to enable such a modularization.

dinosaure:
It's a deep usage of `mimic` where I shifted the whole complexity about protocols and differences between Git, SSH and HTTP outside the core library.

By this way, `HTTP` module is not required to have a minimal implementation of Git and `git_mirage_http` + `git_paf` takes care about the HTTP layer and must be composed if we want to support HTTP(S).

The idea of this PR is to let _via_ functoria the user to choose which protocol he wants to support to synchronize wit a Git repository - and, by this way, have the possibility to craft a smaller unikernel if he does not want to support `HTTP` - so it should fix #513.

Such composition is done by `mimic` where, for any protocols, it returns a new `Mimic.ctx` which should be merge with one from another implementation. Then, Git, SSH or HTTP requires `happy-eyeballs` finally to communicate with IPv4 or IPv6. The PR is still a draft, I need to compose by myself the whole stack but in my head, it seems clear enough.


dinosaure:
- [ ] Fix the `V1` argument

dinosaure:
/cc @hannesm, I think this version is ready! You have different new `devices` which bring a certain protocol. For instance, if you don't want to bring the HTTP(S) implementation, you just need to **not** _merge_ the `mimic_http` device with others (for instance, `mimic_tcp` and `mimic_ssh`).

All of them use `happy-eyeballs` to do the DNS resolution - however, we can not (yet!) set timers on this specific device.

Then, the user is able to set SSH authenticator/TLS key authenticator/TLS certificate authenticator to ensure to talk with the right peer if he wants.

Finally, the core of the Git protocol does not need anymore an HTTP implementation as a _functor_. So I broke the API and `irmin` should be updated with this new interface but, as I said, the usage of these transmission protocols (TCP/IP, SSH or HTTP with or without TLS) is more homogeneous (so we should have less bugs because we have less different execution paths).

PR #544, opened 2021-12-13
Link with logs and logs.fmt

MisterDA:
There are a couple of places where the link with `logs` or `logs.fmt` was forgotten in 3.6.0. This breaks the package installation in my opam switch :/

### DIFF
--- a/bin/guit/dune
+++ b/bin/guit/dune
@@ -8,6 +8,7 @@
   git
   git.nss.git
   logs
+  logs.fmt
   fmt
   mtime
   mtime.clock.os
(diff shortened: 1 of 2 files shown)


dinosaure:
LGTM but how you get the linking problem?

MisterDA:
Just compiling ocaml-git in my switch or installing it as a dependency via `opam install --deps-only .` of my project raises this problem

dinosaure:
> Just compiling ocaml-git in my switch or installing it as a dependency via opam install --deps-only . of my project raises this problem

That weird that the CI did not catch it. Let's merge, thanks!

PR #548, opened 2021-12-17
Upgrade to tcpip.7.0.0 and remove mirage-protocols/mirage-stack

dinosaure:


dinosaure:
The last commit permits to upgrade `docteur` and by this way, we don't lost features for the UNIX backend. By this way, the `Git_unix.ctx` is able to, at least, initiate a FIFO and HTTP(S) connection and it only needs an `Happy_eyeballs_lwt.t`. An `awa-ssh` connection can be initialized as far as the user give some information _via_ `Git_mirage_ssh.with_optional_key`. Of course, `git://` works too.

PR #551, opened 2022-01-04
git: provide a `Store.read_opt` function for finding potentially-absent objects

CraigFe:
Following discussion in https://github.com/mirage/irmin/pull/1698.


dinosaure:
Thanks!

PR #559, opened 2022-02-21
Fix usage of ephemerons

dinosaure:
A possible fix for OCaml 5.0 and ephemerons.

### DIFF
--- a/src/carton/idx.ml
+++ b/src/carton/idx.ml
@@ -555,24 +555,28 @@ end = struct
   let encode e = e.k e
 end
 
+type file = File
+
+module Ephemeron = Ephemeron.K1.Make (struct
+  type t = file
+
+  let equal = ( = )
+  let hash = Hashtbl.hash
+end)
+
 module Device = struct
-  type 'uid value = Bigstringaf.t
-  type key = Key
-  type uid = key ref
-  type 'uid t = (key ref, 'uid value ref) Ephemeron.K1.t
+  type t = Bigstringaf.t ref Ephemeron.t
+  type uid = file
 
-  let device () = Ephemeron.K1.create ()
+  let device () = Ephemeron.create 1
 
-  let create tbl =
-    let key = ref Key in
-    Ephemeron.K1.set_key tbl key;
-    Ephemeron.K1.set_data tbl (ref Bigstringaf.empty);
-    key
+  let create device =
+    let file = File in
+    Ephemeron.add device file (ref Bigstringaf.empty);
+    file
+    [@@inline never]
 
-  let project tbl uid =
-    assert (Ephemeron.K1.get_key tbl = Some uid);
-    match Stdlib.Option.get (Ephemeron.K1.get_data tbl) with
-    | { contents = v } -> v
+  let project device file = !(Ephemeron.find device file)
 end
 
 module M (IO : sig
@@ -595,6 +599,7 @@ struct
     mutable buffer : Bigstringaf.t;
     mutable capacity : int;
     mutable length : int;
+    uid : Device.uid;
   }
 
   let enlarge fd more =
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
@damiendoligez @stedolan @kayceesrk does any of you know a way to avoid uses of `Obj.Ephemeron` here? Could something be added to the API to help this use-case?

dinosaure:
For more precision, I would like to implement "a device of temporary objects" where, in `src/carton/idx.ml`, it's about one object and `src/git/cstruct_append.ml` is 2 objects. And, as long as I keep key given by this "device", values stay alive. The real usage is available here:
https://github.com/mirage/ocaml-git/blob/844087dc77b9650ba8d27cc823134f3ee0720a81/src/git/mem.ml#L476-L499

Where `t_idx` and `t_pack` are these "devices". From them, we are able to create temporary objects (`Cstruct.t` in reality) and `fetch` manipulates them (without the knowledge of how to "finalize"/"delete" them).

dinosaure:
I just push-forced a version with `Ephemeron.K1.Make`. The version seems to work but I would like to check some stressful cases.

stedolan:
Missed this earlier, but I just had a look and I don't see why ephemerons are needed here at all? All of the calls to `create` use the same key (the constant `File`), so I think the ephemeron has the same effect as an ordinary reference?

dinosaure:
> Missed this earlier, but I just had a look and I don't see why ephemerons are needed here at all? All of the calls to create use the same key (the constant File), so I think the ephemeron has the same effect as an ordinary reference?

Yeah, I'm not sure about what I did but my goal is to simulate via ephemerons a kind of temporary file which lives until we keep the `File` somewhere (for instance, into the `fetch` function) and gives the value to another function (which does not know about `File` but only about the value). In my head is like:
```ocaml
  let k, v = create () in
  join do_something (do_something_in_v v) >>= fun () ->
  let k = k in
  (* the GC will delete [k] and [v]. *)
  return ()
```
PS: @stedolan, completely unrelated but I send you an email/slack message, if you have the time to look on it, it will be super useful for me!

stedolan:
> Yeah, I'm not sure about what I did but my goal is to simulate via ephemerons a kind of temporary file which lives until we keep the `File` somewhere (for instance, into the `fetch` function) and gives the value to another function (which does not know about `File` but only about the value). In my head is like:
> 
> ```ocaml
>   let k, v = create () in
>   join do_something (do_something_in_v v) >>= fun () ->
>   let k = k in
>   (* the GC will delete [k] and [v]. *)
>   return ()
> ```

I still don't really understand what's going on here. What might go wrong if you just ignore `k`, by writing something like the following?
```ocaml
  let _, v = create () in
  join do_something (do_something_in_v v) >>= fun () ->
  ...
  return ()
```
The GC will keep the value alive as long as it's in use. Is this not what you want?

dinosaure:
> The GC will keep the value alive as long as it's in use.

It seems that if we ignore `k`, `v` will be deleted at one time (`do_something_in_v` can take a long time). I can test it if you want however, again, I'm not sure about what I did.

damiendoligez:
Coming back to this, trying to understand the code and your informal descriptions, it seems to me that you are trying to do something similar to the handling of pipes by the Unix kernel: a pipe has two ends (reading and writing), as long as someone is holding on to the reading end, other processes can write to the writing end, but when the reading end is closed (garbage-collected) by everyone, then writes to the writing end raise an error. Is this the kind of thing you want to do?

dinosaure:
Hi @damiendoligez, thanks to looking on this problem again. I should have mentioned on this PR the discussion I had with @stedolan. It is clear that the use of ephemeron is wrong and should not be used. I kept its use because I don't have time to really look into it yet but @stedoaln convinced me of a solution without ephemerons that should work.

The behavior is much more close to a temporary file where we want to store temporarily a file to make an analysis (with `mmap`). The result of this analysis allows us to say:
- either we move only this temporary file to a persistent zone
- or we make another file in the persistent zone from what we get in this temporary file

The lifetime of the temporary file becomes complex because of `lwt` where we have to keep it alive at the end of this analysis (which is done concurrently).

In such context, it seems clear that ephemerons are not needed _à priori_. But because of the depth of this mechanism, I have to give myself a whole week to really look into it.

PR #560, opened 2022-03-07
Fix fmt deps for v.exe

maiste:
Hi dear `ocaml-git` maintainers,
Following this PR (https://github.com/mirage/ocaml-git/pull/499), I added `fmt` as a dependency to `v` to make `opam monorepo` work with `ocaml-git.3.8.0` and `irmin` lock file.

### DIFF
--- a/bin/guit/dune
+++ b/bin/guit/dune
@@ -42,6 +42,7 @@
   logs
   logs.fmt
   logs.cli
+  fmt
   fmt.tty
   fmt.cli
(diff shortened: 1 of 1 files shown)


dinosaure:
Thanks, waiting the CI and I will merge it, a release will be done soon to be compatible with OCaml 5.00.0 and prepare few devices for MirageOS - but I need to dig in few other projects.

PR #561, opened 2022-03-15
Fix the deletion of pack files

dinosaure:
Spotted by @patricoferris, it fixes the `Major.reset` and delete properly PACK files.


patricoferris:
Thanks @dinosaure (spotted with @dinakajoy too ^^)!

PR #568, opened 2022-04-11
Remove bigarray-compat and mmap

dinosaure:


hannesm:
this looks nice! :)

hannesm:
eventually the `result` library could be removed as well, but AFAICT that'll be cumbersome due to Lwt_result and `implicit_transitive_deps false`

PR #575, opened 2022-04-27
carton-lwt.opam: remove bigarray-compat

hannesm:
since this only changes the opam dependency, the same could be done to the carton-lwt 0.4.4 release to opam-repository (i.e. no need for a new release).


dinosaure:
Thanks, I will propose a PR on `opam-repository`.

PR #580, opened 2022-07-23
Do not append a leading slash to path

reynir:
I found that I could not fetch the repo http://git.data.coop/halfd/new-website.git. With full debug logs I was able to gather that http://git.data.coop//halfd/new-website.git was requested instead, and gitea treats that as an error and returns 404.

This PR removes the extra leading slash and adds a test case.


reynir:
There are similar errors in `Smart_git.Endpoint.pp` for git protocol and likely generic scheme protocol.

hannesm:
I just run into this issue as well... would be great to have it merged & released :)

dinosaure (approved):

dinosaure:
Thanks, I will cut a release as soon as I can.

Issue #581, opened 2022-08-03
fd leak when reading files from a checkout


emillon:
Hi,

I'm calling the following function in a loop  I'm calling this function in a loop (it's loading a bunch of opam files from a checkout of opam-repository).

```ocaml
    let read_opam root pkg =
      let open Lwt_result.Syntax in
      let* store = Git_unix.Store.v root in
      let* commit_id = Git_unix.Store.Ref.resolve store Git.Reference.master in
      let open Lwt.Syntax in
      let name = OpamPackage.name pkg in
      let path =
        [
          "packages";
          Format.asprintf "%a" pp_name name;
          Format.asprintf "%a" pp_package pkg;
          "opam";
        ]
      in
      let* blob_opt = Search.find store commit_id (`Commit (`Path path)) in
      match blob_opt with
      | None -> assert false
      | Some blob -> (
          let* r = Git_unix.Store.read store blob in
          match r with
          | Ok (Blob b) ->
              let s = Git.Blob.to_string b in
              let opam = OpamFile.OPAM.read_from_string s in
              Lwt.return (Ok opam)
          | _ -> assert false)
```

It runs out of fds. According to `valgrind --tool=none --track-fds=yes`, the files that are not closed are pack files (such as `.git/objects/pack/pack-d73621315d08cdcf1ec645c3145974593a0aa945.pack`).

I would like these fds to be closed once the data is returned, or as an alternative I think that a "fd pool" approach is possible, so that ocaml-git does not open more than N fds at a time, and closes them all at the end of the pool's lifetime.

Thanks!

moyodiallo:
There's is an issue here, opening store each time in loop. You can open once store and use it. 
`let read_opam root pkg store =..`

emillon:
As mentioned in the linked PR, I'm not using ocaml-git anymore so this does not need fixing on my side. @moyodiallo if you have a related issue please open one. Thanks!

PR #590, opened 2022-09-22
Be able to clean pending file-descriptors kept by ocaml-git

dinosaure:
/cc @emillon, this PR should fix your issue #581. Can you confirm that?

### DIFF
--- a/src/git-unix/git_unix.mli
+++ b/src/git-unix/git_unix.mli
@@ -31,4 +31,11 @@ module Store : sig
   include Git.S with type hash = Digestif.SHA1.t
 
   val v : ?dotgit:Fpath.t -> Fpath.t -> (t, error) result Lwt.t
+
+  val close_pack_files : t -> unit Lwt.t
+  (** [close_pack_files t] closes all {!type:Mj.fd}.
+
+      {b NOTE}: Any subsequent use of the [t] has undefined behavior! This
+      function should be registered with [at_exit] to clean pending
+      file-descriptors. *)
 end


moyodiallo:
I tested this PR and it's OK :+1: , just last thing left.
```
diff --git a/src/git-unix/git_unix.mli b/src/git-unix/git_unix.mli
index 248d1fb0e..baa3512e3 100644
--- a/src/git-unix/git_unix.mli
+++ b/src/git-unix/git_unix.mli
@@ -31,4 +31,6 @@ module Store : sig
   include Git.S with type hash = Digestif.SHA1.t
 
   val v : ?dotgit:Fpath.t -> Fpath.t -> (t, error) result Lwt.t
+
+  val close_pack_files : t -> unit Lwt.t
 end
```

We are using once each time the store without having the ability to close the files descriptor, so getting file descriptor leaks.

Having this for our https://github.com/ocurrent/solver-service would be nice.

dinosaure:
So, this PR fixes your issue (#581) and I should merge it as is?

emillon:
I don't use ocaml-git anymore, so I don't need a fix for #581. But I don't think that's right - I was looking to close the fds before the process ends, so `at_exit` would not work.

@moyodiallo might have a different use case in mind, I'll let him comment what was his issue, I'm not sure it's the same.

moyodiallo:
We need this PR because:

We have parent process that each time we have a new opam-repository commit this https://github.com/ocurrent/solver-service/blob/e60cac6acbadbe9a1eb24acb249ca928db8e6318/service/service.ml#L57-L70 function is used, it opens a new store. And the parent process continues to accumulate files descriptors and passing them to the child processes every time a new one is forked. At some point we are forced to restart the service.

@dinosaure

dinosaure:
Thanks, just to confirm that this PR is good enough for you. I will try to cut a release as soon as I can.

PR #594, opened 2022-10-05
Git_mirage_http: the authenticator format description has been incorporated into x509 0.16.2, use this

hannesm:


dinosaure:
Thanks!

PR #600, opened 2022-10-22
ci: Configure CI for NixOS

### LINKED ISSUE #473: > I will try to find a CI about NixOS on the other side.
> I will try to find a CI about NixOS on the other side.

Not sure if it's worth the hassle since you'd need to maintain a mapping from ocaml dependencies to NixOS packages for that in some shape or form. You could also catch this class of failures by having the following before a test run, I think:

```
export HOME="$(mktemp -d)"
set -xe XDG_CONFIG_HOME
```

This would make `git` cease to find its configuration triggering these kind of test execution bugs.

Or we go down a different path: I also could create a temporary global gitconfig for the tests in the NixOS build environment or we do it somewhere in the tests?

_Originally posted by @sternenseemann in https://github.com/mirage/ocaml-git/issues/472#issuecomment-782610255_

Repeatedly, we need to be care about `git commit` which fails when global configuration (`username`, `email`) is missing - which is the case on NixOS. So, we should find a way to check that and avoid any regression then.

Et7f3:
fix #473

> you'd need to maintain a mapping from ocaml dependencies to NixOS packages 

You have opam2nix used in dune repo that convert automatically or you have curated list (by human) in [nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/ocaml-packages.nix)

You have two workflow:
- one low maintenance but not 100% reproductible (enough to reproduce most issue): You import definition already written in Nixpkgs and you update a little bit for your local dev. The con is you don't get lock file.
- another 100% reproductible is with flake like dune you commit a flake.nix and flake.lock

I opted for the solution with lower maintenance. It should catch most issue with NixOS + if a nix user want an old version it can always checkout an old version of nixpkgs (that is 100% reproductible) so it I would opt to avoid double maintenance.

Is it preferable to have all package's check enabled even the failing one or disabling them is fine ? What is mirage policy ?


dinosaure:
Thanks!

Issue #601, opened 2022-11-14
Compatibility with gitea http implementation


reynir:
With #580 I had hoped I could use ocaml-git against a gitea http repository, but it fails with a 401 Unauthorized when doing a POST request at https://git.data.coop/data.coop/slides.git/git-upload-pack. Unfortunately, it loops retrying the request over and over.

I have discovered that
- it works fine with C-git
- even with smart protocol version 1 it works with C-git
- checking the curl trace from C-git I discovered they set headers `content-type`, `accept`, `git-protocol` and `user-agent`
- adding `content-type` and `accept` results in a 200 OK with an empty body

reynir:
I spent quite some time trying to reproduce the curl invocation from C git, but I always got an empty response. I suspect it may be either binary data I copied over incorrectly (not sure that's the case), or a copy/paste mistake (the curl trace is split at every N characters). At that point I was also somewhat exhausted. I will continue trying to reproduce a working curl invocation.

I will as well submit a patch to add a content-type header as Gitea requires that: https://github.com/go-gitea/gitea/blob/dcd9fc7ee894700f702f3847d7d2a41d6a009b7e/routers/web/repo/http.go#L421

reynir:
Thanks to @hannesm I relearnt the difference between `curl --data` and `curl --data-binary` and I could get a response with a body (after fixing newlines that were either escaped as `.` or `\x0A` depending how I got the POST body).

PR #603, opened 2022-11-15
Specify on headers the Smart version of the protocol and the type of the protocol

dinosaure:
This PR is untested but it probably helps to `fetch`/`push` to a server with the Smart protocol 1 (/cc @reynir who spotted the bug with gitea).

### DIFF
--- a/src/not-so-smart/smart_git.ml
+++ b/src/not-so-smart/smart_git.ml
@@ -383,6 +383,21 @@ struct
         | None -> get_transmission r)
     | [] -> None
 
+  let add_unless lst k v = match List.assoc_opt k lst with
+    | Some _ -> lst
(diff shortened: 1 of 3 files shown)


src/not-so-smart/smart_git.ml
```diff
@@ -383,6 +383,21 @@ struct
         | None -> get_transmission r)
     | [] -> None
 
+  let add_unless lst k v = match List.assoc_opt k lst with
```
reynir, line 386: Are the keys assumed to be lower case? Ideally, we should match case insensitive.

reynir (commented):
I will test it now. I have a minor comment.

reynir:
:partying_face: it works! Thanks! This fixes #601.

PR #605, opened 2022-12-07
Be explicit about the format of Smart_git.Endpoint.of_string

### LINKED ISSUE #604: `Smart_git.Endpoint.of_string`: explain the format
https://github.com/mirage/ocaml-git/blob/76e8cc96790ae364c816c89f1e4dc705a6809b3c/src/not-so-smart/smart_git.ml#L112

dinosaure:
Fix #604


reynir (approved):
\o/ thanks, looks good

PR #608, opened 2022-12-19
Try to bounds the stream between the receiver and the PACK decoder

dinosaure:
Already spotted into `multipart_form` but currently, the stream of the PACK file is not bound. The side effect is: `lwt` does not really do the concurrence between the receiving and the decoding and we need to limit how many chunks we can have into our stream. So we use `Lwt_stream.bounded_push`.

The final expectation is to give opportunities to some processes to run even if we started to decode the PACK file. Now, the decoding is highly bound to the receiver (the _socket_) which can block. That mostly means that, by transitivity, the decoder can block too and let opportunities to some other processes.

### DIFF
--- a/src/not-so-smart/fetch.ml
+++ b/src/not-so-smart/fetch.ml
@@ -113,19 +113,21 @@ struct
     >>= function
     | `Close -> return []
     | `Continue res ->
-        let pack ctx =
+        let recv_pack ctx =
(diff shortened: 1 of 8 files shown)


hannesm:
thanks for working on this, and merging it. would you mind to cut a release of git with this PR included?

PR #613, opened 2023-03-22
git-mirage: allow password authentication, as supported by awa 0.2.0

hannesm:
please let me know if it should be done differently


dinosaure:
Thanks!

PR #622, opened 2023-08-15
Fix tests for git 2.41.0+

### LINKED ISSUE #617: Test failure with git 2.41
I’ve observed failure of the tests for `carton-lwt` when using `git` at version 2.41 (2.40.1 worked well):

> fatal: packfile name '…/_build/default/test/carton/index-null' does not end with '.idx'
> ASSERT run ['git' 'index-pack' '--stdin' '-o' '…/_build/default/test/carton/index-null']: exited with 128
> …
> Called from Dune__exe__Test.empty_index in file "test/carton/test.ml", line 300, characters 26-43

metanivek:
- `git index-path` now requires a '.idx' extension for the index file
- `git fsck` has stricter timezone parsing and caught a bug in date calculation in a test when the current timezone is negative

Fixes #617

### DIFF
--- a/test/carton/test.ml
+++ b/test/carton/test.ml
@@ -277,7 +277,7 @@ let empty_index, uid_empty_index =
   let res =
     let open Rresult in
     Bos.OS.Dir.current () >>= fun current ->
-    let dst = Fpath.(current / "index-null") in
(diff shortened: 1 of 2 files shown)


dinosaure:
LGTM, thanks!

PR #623, opened 2023-08-17
Fix raw parsing in middle of buffer

metanivek:
Address issue with `of_raw_with_header` assuming it could read until the end of a given buffer and add a new function, `length_with_header`, that clients can use to inspect length read.

I ran into this while working on testing `irmin-server`. Its protocol (de)serializes blobs in the middle of a buffer, which does not work when using an `irmin-git` store. I have verified that this patch (with appropriate changes in `irmin-git`) fixes the issue for us.

### DIFF
--- a/src/git/value.ml
+++ b/src/git/value.ml
@@ -46,6 +46,7 @@ module type S = sig
   include S.BASE with type t := t
 
   val length : t -> int64
+  val length_with_header : t -> int64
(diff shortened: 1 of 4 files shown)


dinosaure (approved):

dinosaure:
Seems good, thanks for your work, I will try to cut a release as soon as I can.

PR #634, opened 2024-05-08
Adapt to mirage-flow 4.0.0 changes: add a shutdown function

hannesm:
In the tests I was a bit sloppy and defined shutdown as close.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,4 +1,4 @@
-version=0.25.1
+version=0.26.2
 module-item-spacing=compact
 break-infix=fit-or-vertical
 parens-tuple=multi-line-only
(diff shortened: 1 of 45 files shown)


hannesm:
this is ready now, and a merge & release would be great. to allow mirage to use the new git-mirage then, a 3.15.1 release would be great (if it is 3.16.0, we'll need to adapt the bounds in mirage and cut another mirage release).

thanks for all the great work! :)

hannesm:
I updated the unikernels to mirage 4.5.0 :)

I also shaved some more unneeded dependencies.

this is good to be merged and released :)

dinosaure:
Thanks!

PR #638, opened 2024-05-17
opam: require rresult 0.7.0

hannesm:
there are CI failures in https://github.com/ocaml/opam-repository/pull/25893 that I think are fixed by this.

also https://github.com/dinosaure/opam-repository/pull/1 for your opam-repo PR


hannesm:
this I will close.

PR #641, opened 2024-06-10
git-mirage: avoid failwith, use exit 64 (argument error)

hannesm:
The issue I observed was a unikernel that restarted all the time, but its argument had an issue "wrong padding" was the output, but while there was a backtrace, there was no useful information. I tracked it back that git-mirage-ssh got a ssh private key where the padding was bad.

So, this PR does (a) prefix the output with slightly more information and (b) returns exit code 64 (as defined in Mirage_runtime.argument_error) -- this is crucial since albatross then won't restart such a unikernel (that will not be able to succeed a startup, since the error will persist until the command line argument is changed).


dinosaure:
Thanks!

PR #648, opened 2025-01-05
add x-maintenance-intent field to the opam file

hannesm:
if approved / merged, I'm happy to upstream this to the opam-repository


dinosaure:
Thanks!

PR #650, opened 2025-02-10
use dune variants instead of functors

hannesm:
as advertised in title :)


hannesm:
Dear @dinosaure, would you mind to merge and cut a release of git? This would help us to get over the border with the defunctorization work. Thanks a lot.

PR #652, opened 2025-05-14
Report the error

reynir:
I think this is the source of the exception reported here which is not that informative: https://git.robur.coop/robur/opam-mirror/issues/35

This adds the error message which hopefully has more information.

CC @hannesm


hannesm:
thanks. I'm happy to see this, but for the referenced thing, it would even be more useful if we could achieve a stack trace why that is happening... I'm also curious in which situations read_exn is called which are not safe...

PR #653, opened 2025-05-14
Raise instead of lwt fail

hannesm:
based on #652, go a step further and instead of Lwt.fail, we raise the exception -- with the hope we'll see some more useful backtraces (currently starting the unikernel in question)


reynir (approved):

dinosaure:
Thanks!

